PR #2, opened 2015-04-15
Added test data for capability tests used in Java and PHP libraries

bladeSk:


mattheworiordan:
You mention you're not convinced the extra apps should be include, why not?

bladeSk:
You said the tests that use extra app keys are unnecessary bloat, as they test if the server is returning expected data. I have to agree with that and I think tests like this should be a part of test suite for the server, not the client.

mattheworiordan:
@bladeSk I agree, all of these additional keys for the tests are unnecessary.  The second key with the capabilities of:

```json
{ 
  "cansubscribe:*":["subscribe"], 
  "canpublish:*":["publish"],
  "canpublish:andpresence":["presence","publish"]
}
```
should provide sufficient permutations of capabilities to run your tests.  Do you agree?

bladeSk:
Yep, that's right, closing the PR

PR #3, opened 2015-04-16
Fix encrypted data for crypto fixture

rjeczalik:
The ciphertext for '.items | .[1].encrypted.data' was errornously
padded, which caused ably-go tests to fail on encryption tests (the
decryption was unaffected as the padding was removed either way).

The ciphertext was (in bytes):

  0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15

Which got padded to:

  0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 16 16 16 16 16 16 16 16 16 16
16 16 16 16 16

And such data was encrypted in the old version. This CL changes the
encrypted data to the one created for the ciphertext without padding.

### DIFF
--- a/test-resources/crypto-data-128.json
+++ b/test-resources/crypto-data-128.json
@@ -24,7 +24,7 @@
       },
       "encrypted": {
         "name": "example",
-        "data": "HO4cYSP8LybPYBPZPHQOtuB3dfKG08yw7J4qx3kkjxdW0eoZv+nGAp76OKqYQ327",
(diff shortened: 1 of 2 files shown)


kouno:
Related to this PR: https://github.com/ably/ably-go/pull/16/files#r28551433

rjeczalik:
I was wrong with this fix, the test-cases are correctly padded.

I needed to understand that payloads in those test-cases are encoded for binary transports, so if binary payloads happen to be `n * aes.BlockSize` long it needs to be padded by `16`, otherwise if the last byte of the payload is `< aes.BlockSize` then unpadding received message would fail.

I'm closing this issue and addressing the issue in ably-go with https://github.com/ably/ably-go/pull/18.

Sorry for confusion.

PR #4, opened 2015-08-02
Prefix duplicate keys

fealebenpae:
We don't need the values of the comments in the duplicate keys for `post_apps` and `cipher`, and they can actually break some JSON libraries as the JSON specification doesn't really define any behavior with regards to duplicate keys. Since we don't need them to be machine-readable, just prefix with an underscore to avoid key clashes.

### DIFF
--- a/test-resources/test-app-setup.json
+++ b/test-resources/test-app-setup.json
@@ -1,5 +1,5 @@
 {
-  "post_apps": "/* JSON body using in POST sandbox-rest.ably.io/apps request to set up the Test app */",
+  "_post_apps": "/* JSON body using in POST sandbox-rest.ably.io/apps request to set up the Test app */",
   "post_apps": {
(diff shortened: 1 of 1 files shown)


mattheworiordan:
Looks good. @SimonWoolf this should fix the issues you had with Elixir.  You happy for me to merge in?

SimonWoolf:
Sure. (Though I'll still need to use my own branch because of the channel name changes)

mattheworiordan:
@SimonWoolf I will merge in, but looking at bfee8caacd4e683a592b50b39ed065c73af3b5c2, why not add presence as a namespace as opposed to replace persisted, and then it should be unlikely to impact any of the existing client libraries?

SimonWoolf:
@mattheworiordan makes sense, will do

PR #5, opened 2015-08-28
Add invalid client id code

SimonWoolf:


paddybyers:
oops, I had pending commits which are now pushed, so I added this as well ... but the code is now `40012`.

PR #7, opened 2015-09-03
Fix encoding string in fixture

SimonWoolf:
See https://github.com/ably/ably-js/pull/118


paddybyers:
@SimonWoolf I'll leave you to merge this once you've fixed any associated regressions in ably-ruby

PR #8, opened 2015-09-04
Add some more keys needed for ably-js capability tests

SimonWoolf:


SimonWoolf:
@mattheworiordan any objection?

mattheworiordan:
Nope, no objection ;)

PR #10, opened 2015-10-01
Add msgpack to the crypto fixture data

SimonWoolf:
Prereq for addressing https://github.com/ably/wiki/issues/22

@mattheworiordan @paddybyers any objections?

### DIFF
--- a/test-resources/crypto-data-128.json
+++ b/test-resources/crypto-data-128.json
@@ -14,7 +14,8 @@
         "name": "example",
         "data": "HO4cYSP8LybPYBPZPHQOtmHItcxYdSvcNUC6kXVpMn0VFL+9z2/5tJ6WFbR0SBT1xhFRuJ+MeBGTU3yOY9P5ow==",
         "encoding": "utf-8/cipher+aes-128-cbc/base64"
-      }
+      },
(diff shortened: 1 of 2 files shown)


paddybyers:
> any objections?

No, but right now we don't know how you're going to use these in tests.

It's not valid, unfortunately, to assume key order in the msgpack serialisation, so you can't do an `encode()` and compare the text output.

Verification of the `decode()` can easily be by decoding the msgpack string and doing a deepEqual on the result.

Verification of the `encode()` strictly would need us to have an independent and valid decoder to decode the output of the client library, and then do a deepEqual on the results.

SimonWoolf:
> It's not valid, unfortunately, to assume key order in the msgpack serialisation,

That's annoying. A combination of decoding the fixture and comparing with `deepEqual`, plus comparing the results of msgpack-encoding the `testMessage` and the `encryptedMessage`, would hopefully be enough to catch most bugs (the latter is enough to catch the bug we just fixed), but still, would have been nice to compare the result of `encode()` directly. Oh well.

SimonWoolf:
(Aaand I've just realised all the weird behaviour I've been debugging for the past hour is because I generated these without passing `sparse: true`, and the msgpack decoder can't handle keys with `undefined` values. *headdesk*. Will regenerate)

PR #14, opened 2016-08-03
authUrl and authCallback failure error code

mattheworiordan:
@paddybyers a quick attempt an error code relating to a connection attempt failing due to the auth provider failing.  See https://github.com/ably/docs/pull/132

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -88,6 +88,7 @@
 	"80016": "operation on superseded transport",
 	"80017": "connection closed",
 	"80018": "invalid connection id (invalid format)",
+	"80019": "client configured authentication provider request failed",
 	"80020": "maximum connection message rate exceeded (publish)",
 	"80021": "maximum connection message rate exceeded (subscribe)",
 	"80030": "client restriction not satisfied",


SimonWoolf:
Copying my comment from that PR: there is already a code for this - `"40170": "error from client token callback"`. But actually I think you're right that it should be an 8xxxx error and not a 4xxxxx, so happy to change to 80019. (Should probably remove the 40170 one too)

paddybyers:
@SimonWoolf I was about to make the same comment, so lets remove 40170.

Otherwise LGTM.

mattheworiordan:
I am not convinced we should remove 40170.  For example, if one explicitly called `authorise` and it failed, but it was not related to a connection attempt, would 40170 not be more appropriate?

SimonWoolf:
> I am not convinced we should remove 40170. For example, if one explicitly called authorise and it failed, but it was not related to a connection attempt, would 40170 not be more appropriate?

I can see where you're coming from, but IMO not worth having two different codes for what's basically the same thing, s'just more complexity. (But I don't feel strongly about it, so happy to go with two codes if Paddy agrees with you)

mattheworiordan:
Not overly bothered, I will let @paddybyers decide

SimonWoolf:
@paddybyers you were asking in slack why 80019 wasn't in ably-common -- unmerged PR :)

Reading this issue again now, I actually ended up implementing Matt's preferred solution in ably-js (ie two different error codes -- 40170 for when it's a fatal error, e.g. for a direct `requestToken()` or REST call where there's no retry, and 80019 for when it's part of a connection attempt and is nonfatal as it just retries again after a disconnectedTimeout), so I'll withdraw my suggestion and just merge immediately.

PR #15, opened 2016-08-06
Channel operation timed out error for (RTL3d)

mattheworiordan:
@paddybyers are you happy for me to add this error code that I can emit when a channel attach/detach operation fails? See https://github.com/ably/docs/commit/a135a4009ef93b66c042dc73c8e9ff5a2ce033c8#diff-11900b395df266bc2bbfc1d11bdfc7a0R394


paddybyers:
Adding a code is good. I'd prefer that we make it clear that the timeout is because the server didn't respond, eg `channel operation failed (no response from server)`

mattheworiordan:
@paddybyers merged in

PR #17, opened 2016-08-08
Expected value proposal

paddybyers:
Any comments/updates before we merge?

Perhaps `expectedHexValue` for the binary case?

I still think `object` and `array` would have been fine :)

### DIFF
--- a/test-resources/messages-encoding.json
+++ b/test-resources/messages-encoding.json
@@ -3,7 +3,7 @@
   "messages": [
     {"data": "foo", "encoding": null, "expectedType": "string"},
     {"data": "{\"foo\":42}", "encoding": null, "expectedType": "string"},
-    {"data": "{\"foo\":42}", "encoding": "json",  "expectedType": "map"},
+    {"data": "{\"foo\":42}", "encoding": "json",  "expectedType": "object"},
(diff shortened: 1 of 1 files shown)


mattheworiordan:
> Perhaps expectedHexValue for the binary case?

I wouldn't be against that

> I still think object and array would have been fine :)

I disagree.  I like the fact that we are being explicit that `object` and `array` are as defined in the JSON spec for `object` and `array` as opposed to generic `object` and `array` implementations.

tcard:
I actually went ahead and cherry-picked your commit onto https://github.com/ably/ably-common/pull/16.

Kind of messy that we have three PRs now overlapping and/or building on top of each other. Should I close this, merge #16 (which incorporates this) and then review/merge #18 ?

tcard:
Actually, I've closed #16 and now I think we should merge this one and then #18. Both look good to me and I'll go ahead an update https://github.com/ably/ably-ios/pull/459 to #18.

mattheworiordan:
I agree.  We should prioritise this as it is blocking our PHP dev as he does not want to reference a branch.  @paddybyers you ok to do what @tcard suggested?

paddybyers:
> @paddybyers you ok to do what @tcard suggested?

I merged them both without squashing.

PR #18, opened 2016-08-08
Expected value proposal update

mattheworiordan:
Building on @tcard and @paddybyers PRs, I propose we:

* Include a more complex JSON type with objects and arrays inside it
* Be explicit about objects and arrays being JSON objects and arrays as defined in https://tools.ietf.org/html/rfc7159#page-6

### DIFF
--- a/test-resources/messages-encoding.json
+++ b/test-resources/messages-encoding.json
@@ -29,7 +29,7 @@
     	"data": "3q2+7w==",
     	"encoding": "base64",
     	"expectedType": "binary",
-    	"expectedValue": "deadbeef"
+    	"expectedHexValue": "deadbeef"
     }
   ]
 }
--- a/test-resources/messages-encoding.json
+++ b/test-resources/messages-encoding.json
@@ -16,13 +16,13 @@
     {
     	"data": "{\"foo\":42,\"bar\":[\"a\",1.2,{\"boo\":\"ha\"}]}",
     	"encoding": "json",
-    	"expectedType": "object",
+    	"expectedType": "jsonObject",
     	"expectedValue": {"foo":42}
     },
     {
     	"data": "[{\"foo\":42},\"qux\"]",
     	"encoding": "json",
-    	"expectedType": "array",
+    	"expectedType": "jsonArray",
     	"expectedValue": [{"foo":42}, "qux"]
     },
     {
--- a/test-resources/messages-encoding.json
+++ b/test-resources/messages-encoding.json
@@ -14,7 +14,7 @@
     	"expectedValue": "{\"foo\":42}"
     },
     {
-    	"data": "{\"foo\":42}",
+    	"data": "{\"foo\":42,\"bar\":[\"a\",1.2,{\"boo\":\"ha\"}]}",
     	"encoding": "json",
     	"expectedType": "object",
     	"expectedValue": {"foo":42}
--- a/test-resources/messages-encoding.json
+++ b/test-resources/messages-encoding.json
@@ -17,7 +17,7 @@
     	"data": "{\"foo\":42,\"bar\":[\"a\",1.2,{\"boo\":\"ha\"}]}",
     	"encoding": "json",
     	"expectedType": "jsonObject",
-    	"expectedValue": {"foo":42}
+    	"expectedValue": {"foo":42, "bar":["a", 1.2, {"boo":"ha"}]}
     },
     {
     	"data": "[{\"foo\":42},\"qux\"]",


mattheworiordan:
@paddybyers @tcard any thoughts on whether these interop tests should be extended to MsgPack?

tcard:
@mattheworiordan The JSON/MsgPack encoding of _protocol_ messages works at a different level. We can also test it with fixtures, but that would be unrelated to this.

I don't think it would add much value as it is implicitly tested just by communicating with the realtime server, which, as opposed to message data encoding which go unnoticed by realtime, decodes and encodes protocol messages, so every test would break if the client library did it wrong.

mattheworiordan:
> We can also test it with fixtures, but that would be unrelated to this.

Yup, I sort of agree.  I realise that unlike our encoding for JSON, MsgPack typically handles all the types natively so we're buying into MsgPack's ability to encode & decode.

I'd be happy to stick with what we have for now.

paddybyers:
We have had msgpack interop issues relating to handing of strings vs binary, especially as older msgpack versions didn't really make a good distinction between them.

Remember also that we need to encode and decode a `Message` using msgpack for the REST API.

So I think in this fixture data have various `Message` structs with binary and string `data` members and the corresponding msgpack binary as an `expectedHexValue`.

I would be satisfied that if that works, then the corresponding encode/decode of a `ProtocolMessage` is as expected, assuming the library tests also test interop with the realtime system, and I would expect those tests to include publish + history retrieval of both string and binary data.

mattheworiordan:
Ok, so are you proposing we add a new `expectedMsgPackHexValue` field so that we publish using MsgPack and check result in JSON and vice versa?

paddybyers:
> Ok, so are you proposing we add a new expectedMsgPackHexValue field so that we publish using MsgPack and check result in JSON and vice versa?

Well at the risk of changing everything we agreed we could include the expected serialised representations of each message in the fixtures with something like:

```
"messages": [
	{
		message: {
			"data": "foo",
			"encoding": null,
			"expectedType": "string",
			"expectedValue": "foo"
		},
		serialised: {
			json: "...",
			msgpackHex: "...
		}
	},
	{
		...
	}
]
```

As for how we test these, I think each library could include tests that use the serialise/deserialise primitives in the library to test.

It's not going to work to be ok to publish and get history with the rest API and check the raw serialised text returned because there will be fields added such as `id` and `timestamp`.

mattheworiordan:
> Well at the risk of changing everything we agreed we could include the expected serialised representations of each message in the fixtures with something like:

I am not against that.  Now is the time to get this right.

> As for how we test these, I think each library could include tests that use the serialise/deserialise primitives in the library to test.

Perhaps, but I would argue we do still need an end-to-end test as well as I don't like testing the inner workings of a library.  End to end could be done perhaps just one way when using the complete serials message body, and then use JSON for the existing publish & subscribe tests as they are. 

WDYT @paddybyers @tcard?

tcard:
I like what @paddybyers proposes, but I don't see what end-to-end buys us in this case.

mattheworiordan:
> I like what @paddybyers proposes, but I don't see what end-to-end buys us in this case.

Well in the same vein as all other tests, we prefer tests that communicate with Ably directly as opposed to testing the internals.  If we test that encoders & decoders give us the result we want, but don't actually test that against the realtime system, there is a chance the HTTP library / WebSocket transport will manipulate the payload.  I am not sure why we need to special case MsgPack tests as internal when all other tests aim to use Ably directly.

paddybyers:
> I don't like testing the inner workings of a library

I agree with the principle, but we do already do it with the msgpack element of the crypto fixtures and encode/decode tests.

> End to end could be done perhaps just one way when using the complete serials message body

I'm not really sure what you're proposing. However, we can include the fixture data and that, by itself, doesn't mandate only one way of using that data in the library tests.

I propose modifying the above to bring it into line with the existing msgpack encoding data in https://github.com/ably/ably-common/blob/master/test-resources/crypto-data-128.json, so we would have:

```
"messages": [
	{
		"message": {
			"data": "foo",
			"encoding": null,
			"expectedType": "string",
			"expectedValue": "foo"
		},
		"msgpack":  "..(base64).."
	},
	{
		"message": { ... },
		"msgpack":  "..(base64).."
	},
	{
		...
	}
]
```

tcard:
In fact, does this even make sense given that the same message can encode to different MsgPack blobs e. g. because of order of fields?

mattheworiordan:
> I agree with the principle, but we do already do it with the msgpack element of the crypto fixtures and encode/decode tests.

You may do, [I mostly don't](https://github.com/ably/ably-ruby/blob/master/spec/acceptance/rest/message_spec.rb#L152-L179)

> I'm not really sure what you're proposing. However, we can include the fixture data and that, by itself, doesn't mandate only one way of using that data in the library tests.

I know, but I have previously proposed that once this is finalised we'll add new spec items to the 0.8 spec.

> I propose modifying the above to bring it into line with the existing msgpack encoding data in

I'm not quite sure what that format is.  Can you explain?

> In fact, does this even make sense given that the same message can encode to different MsgPack blobs e. g. because of order of fields?

Good point :(  Not sure of a practical solution there, we could have deep comparisons of types, but that feels like fixture data will not do the job very well

paddybyers:
> does this even make sense given that the same message can encode to different MsgPack blobs e. g. because of order of fields?

Well, no, it doesn't :)

However, although JSON explicitly says an object is an unordered collection, in practice most JSON libraries will enumerate fields in insertion order, so a test that constructed the object from the given JS literal, and then serialised it, would get the same field order as the given encoded value.

In the equivalent ably-js tests using the crypto data, it doesn't compare the binary encoded values; it decodes the given msgpack data and does a `deepEqual()` on the resulting objects - obviously this is testing the msgpack decode only.

To test msgpack encode then you need to send the encoded data so something to do the decode - so I suppose that might as well be realtime (ie do a rest publish + history test), or an endpoit we provide that does a decode+echo in order to support running the tests.

PR #21, opened 2017-04-05
Define categories for 500 errors

paddybyers:
We will be defining more specific 500 error codes; this sets out a categorisation for them.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -61,6 +61,26 @@
 	"50002": "internal connection error",
 	"50003": "timeout error",
 
+	/* frontend connection-related */
+	"55000": "internal error (uncategorised)",
+
+	/* frontend connection-related */
+	"55100": "internal error (generic connection error)",
+
+	/* frontend channel-related */
+	"55200": "internal error (generic channel or attachment error)",
+	"55201": "internal error (channel rpc error)",
+
+	/* core channel-related */
+	"55300": "internal error (generic message processing error)",
+	"55301": "internal error (core rpc error)",
+
+	/* core webhook/reactor-related */
+	"55400": "internal error (generic reactor error)",
+
+	/* persistence-related */
+	"55500": "internal error (generic persistence error)",
+
 	/* webhook-related */
 	"70000": "webhook operation failed",
 	"70001": "webhook operation failed (post operation failed)",
--- a/test-resources/test-app-setup.json
+++ b/test-resources/test-app-setup.json
@@ -5,7 +5,7 @@
     "keys": [
       {},
       {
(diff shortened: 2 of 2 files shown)


protocol/errors.json
```diff
@@ -61,6 +61,26 @@
 	"50002": "internal connection error",
 	"50003": "timeout error",
 
+	/* frontend connection-related */
+	"55000": "internal error (uncategorised)",
+
+	/* frontend connection-related */
```
mattheworiordan: 55000 and 55100 are both connection related?

protocol/errors.json
```diff
@@ -61,6 +61,26 @@
...
+	"55201": "internal error (channel rpc error)",
+
+	/* core channel-related */
+	"55300": "internal error (generic message processing error)",
+	"55301": "internal error (core rpc error)",
+
+	/* core webhook/reactor-related */
+	"55400": "internal error (generic reactor error)",
+
+	/* persistence-related */
```
mattheworiordan: Would this include History persistence, or persistence more generally?

protocol/errors.json
mattheworiordan: What about auth failures?

protocol/errors.json
mattheworiordan: Should we also include specific error codes for frontends & cores for siege mode i.e. busy?

protocol/errors.json
paddybyers: > 55000 and 55100 are both connection related?

No, the comment on 55000 is wrong.

paddybyers: > What about auth failures?

We could add a category for auth but the internal errors I can think of would fit into other categories: eg persistence failure for persistent token; persistence failure for redis lookup failure; core rpc error for failure to instance an app client.

protocol/errors.json
paddybyers: > Would this include History persistence, or persistence more generally?

It would be generally, and I thinking of errors in the cassandra/redis layers where you don't know whether you're doing something for history or something else.

paddybyers: > Should we also include specific error codes for frontends & cores for siege mode i.e. busy?

Initially I wasn't expecting these to be 500 errors tbh; other rate-limiting errors are 4xx. But I think the existing siege mode functionality we added has 500.

mattheworiordan: > Initially I wasn't expecting these to be 500 errors tbh; other rate-limiting errors are 4xx. 

But they are.  It's not a client problem, it's an Ably service problem, so they are server errors i.e. 5xx.

QuintinWillison:
This pull request does not have a linked issue, the opening explanation doesn't supply enough context, it has unresolved conversations and a search across Ably org GitHub repositories doesn't yield any evidence that these codes are in use anywhere. Hence, I'm closing.

PR #27, opened 2018-05-03
Add errorcodes for exchange access

SimonWoolf:


protocol/errors.json
```diff
@@ -78,6 +78,16 @@
 	"70002": "reactor operation failed (post operation returned unexpected code)",
 	"70003": "reactor operation failed (maximum number of concurrent in-flight requests exceeded)",
 
+	/* exchange-related */
```
paddybyers: TBH I think we could space these out a bit; eg

71000: Product error (unspecified)
71001: No such product
71002:  Product disabled
....

72000: Subscription error (unspecified)
72001:  No such subscription
72002:  Subscription disabled
...

SimonWoolf: @paddybyers how about d8b5a42 ?

paddybyers:
LGTM, thanks

PR #29, opened 2018-07-24
Add default value for maxMessageSize

funkyboy:
So that we can assume this value as default for tests.

### DIFF
--- a/test-resources/test-app-setup.json
+++ b/test-resources/test-app-setup.json
@@ -3,7 +3,7 @@
   "post_apps": {
     "limits": { 
                 "presence": { "maxMembers": 250 },
-                "messages": { "maxMessageSize": 16384}
+                "messages": { "maxSize": 16384}
               }
     "keys": [
       {},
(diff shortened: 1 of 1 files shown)


funkyboy:
@SimonWoolf thanks for spotting this. Fixed here https://github.com/ably/ably-common/pull/29/commits/ab126a4efa3f3c57e1a17bfcdd2660a4ac333a87

SimonWoolf:
assuming you've now tested it (with a value different from the default so you can see whether it's actually worked or not), :+1:

funkyboy:
@SimonWoolf can't test thoroughly until this is deployed to sandbox but the test is already ready: https://github.com/ably/ably-ios/pull/742/files#diff-8c3eccfb7635c28b9fadd0e4fcb6e6b0R47

SimonWoolf:
> can't test thoroughly until this is deployed to sandbox

The ably-common repo is not deployed anywhere. It's pulled in by client libraries.

paddybyers:
It looks to me that realtime doesn't pay any attention to the specified limits in the app spec

mattheworiordan:
@SimonWoolf @paddybyers I am closing this PR as it seems we've not agreed anyway it should be merged, and additionally because it's invalid JSON anyway.  Feel free to reopen if you want.

PR #41, opened 2018-11-20
New error codes for edge caching service failures not served by the Ably platform product

mattheworiordan:
For 502, 503 and 504 error codes, it's likely these errors won't come from the Ably realtime product, but will instead come from routing layers in between.  These error codes will be shown to users by the edge server with a custom error page when a 502, 503 or 504 is served.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -79,10 +79,12 @@
 	"50002": "internal connection error",
 	"50003": "timeout error",
 	"50004": "Request failed due to overloaded instance",
-	"51000": "Ably's edge proxy service has encountered an unknown internal error whilst processing the request",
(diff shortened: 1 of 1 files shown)


paddybyers:
I'm not sure how much this matters, but the 510xx codes would ordinarily be there to supply additional information for a 510 status code. Should we use codes that relate to the underlying status code, eg

```
	"50310": "Ably's edge proxy service received a service unavailable response code from the Ably platform",
	"50410": "Ably's edge proxy service timed out waiting for the Ably platform",
```

?

mattheworiordan:
Yes I’d prefer to do that. I thought you often aimed to define ranges for related error codes, so assumed you’d prefer this. I definitely prefer your codes if you’re happy with that.

mattheworiordan:
@paddybyers FYI 
0e201aa

PR #42, opened 2019-01-07
Test JSON file for use in RealtimePresenceTest

amsurana:
I have created a Test JSON Array file that is used to test PresenceMessage.fromEncodedArray(...) method, which is part of Feature Spec TP4.


mattheworiordan:
@paddybyers will you review please?  I don't have the background I am afraid.  Why can't we simply use https://github.com/ably/ably-common/blob/master/test-resources/messages-encoding.json?

paddybyers:
> Why can't we simply use https://github.com/ably/ably-common/blob/master/test-resources/messages-encoding.json?

These contain an `action`. Since `PresenceMessage.action` is not nullable, it shouldn't be possible to instance a `PresenceMessage` without an `action` specified.

PR #43, opened 2019-01-08
Support articles for error codes 50000, 50003 and 40001

Srushtika:


tomczoink:
@Srushtika LGTM, I edited in links to the debugging article in two of them, so if you're happy with that it's good to merge 👍 .

PR #44, opened 2019-01-28
Active Traffic Management error code

mattheworiordan:
This will be delivered along with an HTTP status code of 503, which will trigger clients to attempt to connect to fallback host.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -80,12 +80,15 @@
 	"50002": "internal connection error",
 	"50003": "timeout error",
 	"50004": "Request failed due to overloaded instance",
-
+	
 	/* Edge cacheing / proxy service error codes */
 	"50010": "Ably's edge proxy service has encountered an unknown internal error whilst processing the request",
 	"50210": "Ably's edge proxy service received an invalid (bad gateway) response from the Ably platform",
 	"50310": "Ably's edge proxy service received a service unavailable response code from the Ably platform",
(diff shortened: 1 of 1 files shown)


mattheworiordan:
Excuse the whitespace btw. I used the Github editor to save time :/

protocol/errors.json
```diff
@@ -80,12 +80,15 @@
...
-
+	
 	/* Edge cacheing / proxy service error codes */
 	"50010": "Ably's edge proxy service has encountered an unknown internal error whilst processing the request",
 	"50210": "Ably's edge proxy service received an invalid (bad gateway) response from the Ably platform",
 	"50310": "Ably's edge proxy service received a service unavailable response code from the Ably platform",
 	"50410": "Ably's edge proxy service timed out waiting for the Ably platform",
+	
+	/* Active Traffic Management error code to indicate intentional redirect of traffic to fallback hosts */
+	"51001": "Active Traffic Management: traffic for this cluster is being temporarily redirected to a backup service",
```
paddybyers: This code doesn't have anything to do with `510 Not Extended (RFC 2774)`. I think 503 (Service Unavailable) is closer. Could we use 50320 ?

mattheworiordan: Ok, sure, I thought for some reason we had abandoned trying to always keep error codes inline with status codes.  I'll update it.

PR #46, opened 2019-05-22
Added 40012, 80016, 40171, 40012, 90001, 40161 to support links

tomczoink:


protocol/errorsHelp.json
```diff
@@ -1,28 +1,32 @@
 {
-  "40000": "https://support.ably.io/solution/articles/3000082819-error-code-40000-bad-request",
```
mattheworiordan: @tomczoink you shouldn't have mixed a commit that changed the URLs for existing error codes, with a commit that added new error codes.  It's harder to review, and harder someone down the line to see what this commit was actually about.

mattheworiordan:
I understand you can’t review articles here btw, but equally I am not aware of any way Tom who wrote the articles can see what you’ve changed in Freshdesk. As such, we’re missing some opportunities to learn as we go along. Anyway thanks all for reviewing

PR #49, opened 2019-12-02
New 256 bit AES CBC encrypted data generated by Java client library SDK

QuintinWillison:
Not really a huge amount for reviewers to look at here, so more informational as I'm adopting these fixtures across the libraries that get heavier utilisation. There's a 'plain text' message here for every byte count from 1 to 70, inclusive.

Generated by the `encryptAndDecrypt()` method in [`CryptoTest.java`](https://github.com/ably/ably-java/blob/531-encryption-investigations/lib/src/test/java/io/ably/lib/util/CryptoTest.java).


QuintinWillison:
I was not planning to deprecate the older files as they have a different focus (e.g. they include content that's not solely binary). More tests, the merrier, or something like that!

mattheworiordan:
Why didn't you just extend `crypto-data-256` instead?  A new file will require a change to the spec https://docs.ably.io/client-lib-development-guide/features/#RSL5c, which I assume is unnecessary?

QuintinWillison:
I simply wrote code that generated a single, atomic document ready to feed in to other client libraries. The fixture file format just seemed to be an appropriate and convenient carrier, from my perspective as someone new to our codebases.

I didn't imagine, for a second, that these files would actually be referenced from the spec. To me that seems a little bit over the top, as surely the spec's role should be to simply define concisely the format of the encrypted messages? How they are tested is down to how much effort has gone in to the library in question, or was I wrong in that interpretation?

paddybyers:
The fixtures are referenced from the spec in https://docs.ably.io/client-lib-development-guide/features/#RSL5c.

mattheworiordan:
> To me that seems a little bit over the top, as surely the spec's role should be to simply define concisely the format of the encrypted messages?

Agreed completely in theory.

In practice though, we have integration tests not because we feel the spec isn't clear, but because bugs exist, misunderstandings happen, and platform inconsistencies can result in unexpected behaviours.  The purposes of the fixtures is to address all of those points.

That does not mean we should not be continuing to improve the spec, we do, but ultimately extensive test coverage + a tightly defined spec gives us the best chance of error-free software for our clients IMHO.  I have in fact continue to advocate that our spec should, where possible, be explicit about what types of tests would be useful to prove the spec item has been implemented correctly.  This is primarily because the person thinking about how to write the spec is deep in the technical detail behind it, and arguably is best positioned at that time to think about what type of tests would "prove" the spec works under not just the happy path, but other paths too.  It's hard for a developer following a spec to think in the same way IMHO.

QuintinWillison:
Thanks both for your input. I'm going to append my fixtures to the existing files.

QuintinWillison:
@mattheworiordan / @paddybyers - any chance that you could cast your eye over this PR one more time? I would like to 'squash and merge' as soon as possible, creating a new commit in `master`, so that I can then update the submodule pointers in the client library SDK repositories.

paddybyers:
Did you plan also to do the 128-bit versions?

QuintinWillison:
Not for now, no. We know that the issue being investigated is with 256 bit and I was also keen to move on to other things! Is that ok, @paddybyers ?

paddybyers:
> Is that ok

I'm ok with leaving it if we raise an issue to get that done at a later time.

QuintinWillison:
Thanks. Done w/ #50

PR #53, opened 2020-02-14
add grpc error codes

cruickshankpg:
There were some missing error codes used by the grpc code defined in realtime's grpcserviceconfiguration.js

```
GRPCServiceConfiguration.GONE = 41000;
GRPCServiceConfiguration.CONNECT_TIMEOUT = 410010;
GRPCServiceConfiguration.NOTHERE = 31000;
GRPCServiceConfiguration.SUPERSEDED = 32000;
```

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -2,6 +2,10 @@
 	/* generic error codes */
 	"10000": "no error",
 
+	/* 310 codes */
+	"31000": "not here",
+	"32000": "superseded",
+
 	/* 400 codes */
 	"40000": "bad request",
 	"40001": "invalid request body",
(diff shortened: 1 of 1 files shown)


SimonWoolf:
@cruickshankpg errors.json in this repo is not supposed to be an exhaustive list of all error codes we use, it's supposed to be an exhaustive list of all error codes _that customers might get_ -- it's a public repo, which is pulled in as a submodule by most clients libraries. (For example, note how the whole [6xxxx](https://github.com/ably/ably-common/blob/master/protocol/errors.json#L96) namespace is just described as 'reserved for internal use', we don't list the internal-use codes here). There are plenty of error codes used for communicating between different bits of the system that would not be sent to a client, and those should not be included in this repo -- I think the ring state error codes you have here fall under this.

(Which does leave open the question of where they should be. If you'd like them to be shared between realtime and go-services, then https://github.com/ably/ably-realtime-common/ is probably the appropriate place)

cruickshankpg:
agree these shouldn't live here

PR #55, opened 2020-05-01
Add 41000 for device registration expired

cruickshankpg:


protocol/errors.json
```diff
@@ -72,6 +72,9 @@
 	/* 405 codes */
 	"40500": "method not allowed",
 
+	/* 410 codes */
+	"41000": "push device registration expired",
```
paddybyers, line 76: We've tended to leave the `00` codes unqualified so they just generically represent the status code.

So perhaps we'd have

```
41000 resource gone (unspecified)
41010 push device registration expired
```

wdyt?

cruickshankpg: yeah makes sense

Issue #32, opened 2018-08-31
Include error constants in includes in this repo?


paddybyers:
Instead of expecting the libraries all to use explicit numeric constants for error codes, we can generate constants for them, and have include files in this repo; for example for go:

```
package ably

// ably error codes
const (
	ErrNothing                                                       = 10000
	ErrBadRequest                                                    = 40000
	ErrInvalidRequestBody                                            = 40001
	ErrInvalidParameterName                                          = 40002
	ErrInvalidParameterValue                                         = 40003
	ErrInvalidHeader                                                 = 40004
	ErrInvalidCredential                                             = 40005
	ErrInvalidConnectionID                                           = 40006
	ErrInvalidMessageID                                              = 40007
	ErrInvalidContentLength                                          = 40008
	ErrMaximumMessageLengthExceeded                                  = 40009
	ErrInvalidChannelName                                            = 40010
	...
	ErrMemberImplicitlyLeftPresenceChannelConnectionClosed           = 91100
)
```

This was generated by a script, and could easily also be done for any other language.

I propose we add this file at `<repo>/protocol/include/errors.go`, add the script somewhere, and add the other includes as and when we're updating the relevant libraries.

@mattheworiordan @SimonWoolf wdyt?

cc @gernest

paddybyers:
It has been pointed out that this would require the submodule to be present for users just getting and building the library (ie not just library developers), so perhaps too much of an inconvenience.

Perhaps just add the script here, not the include file, so other libs can generate a consistent set of constants in their own language.

mattheworiordan:
See https://github.com/ably/ably-ruby/pull/171.  I think it's nice, and my approach does not require access to the errors JSON file at runtime.  However, in untyped languages there is a risk things could break badly if the wording of an error code is changed.

SimonWoolf:
> there is a risk things could break badly if the wording of an error code is changed.

Might be better have each entry in errors.json be an object with separate `name` and `description` fields, rather than try and autogenerate names from the description (meaning we can never update/clarify a human-readable description once it's been added)? Bonus: the error log could source the latest description automatically.

Definitely agree that we shouldn't require access to the submodule at runtime. @mattheworiordan's approach of having a task in the client lib repo to generate (and commit) a language-specific errorcode file using the submodule (and so make sure the result passes ci before releasing it) seems like the right approach :+1:

gernest:
[...]

PR #56, opened 2020-06-04
errors.json: add 40019 and 40021

SimonWoolf:
cf https://github.com/ably/ably-js/pull/658


SimonWoolf:
> Any immediate impact on client libraries now? Or perhaps just when we magic up #32?

Exactly -- pending 32, nothing in errors.json has any direct impact on client libraries yet

PR #57, opened 2020-06-14
Add error codes for push registration operations

paddybyers:


QuintinWillison:
This pull request does not have a linked issue, does not have any explanatory commentary and a search across Ably org GitHub repositories doesn't yield any evidence that these codes are in use anywhere. Hence, I'm closing.

PR #60, opened 2020-11-27
Add 42200 Unprocessable Entity code

twist900:
Support `422 Unprocessable Entity` errors.

> the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -78,6 +78,9 @@
 	/* 410 codes */
 	"41001": "push device registration expired",
 
+	/* 422 codes */
+	"42200": "Unprocessable entity",
+
 	/* 429 codes */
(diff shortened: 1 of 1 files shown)


SimonWoolf (changes requested):
Sorry for the delay, this slipped through the cracks.

Generally the convention this documented uses for codes which add no information on top of the underlying http status code is to append `00` to that, e.g. 40100, 40300, 40400, etc. Any additional information on top of that uses the last two digits. So by that convention this would be 42200 rather than 40022.

SimonWoolf (approved):

PR #62, opened 2021-01-20
Update errorsHelp.json

markcoxsmith:
Updating to remove the broken links. 40010 still doesn't exist for some reason - currently looking


mattheworiordan (approved):
Nice! 👍

kennethkalmer (changes requested):
@markcoxsmith from what I can see `40010` means the channel name is not valid. Could we get a page created in the knowledgebase for this and then have this updated?

kennethkalmer (approved):

paddybyers:
@markcoxsmith For future reference, please familiarise yourself with https://ably.atlassian.net/wiki/spaces/ENG/pages/312574074/Git+Commit+Best+Practice and https://ably.atlassian.net/wiki/spaces/ENG/pages/312377505/Git+PR+best+practice. In this case the commit message isn't helpful - it should indicate which feature/function/category is the subject of the change, and should contain some information about the nature of the change. In this case, it might have been "errors.json: update to new knowledgebase URLs" or something.

Also, please use branch names in line with the policy linked. In this case I would have had `feature/update-knowledgebase-urls` or something like that.

markcoxsmith:
Thanks for the feedback, Paddy. Will be giving these a read through and taking this onboard for the future

PR #63, opened 2021-01-25
Rake task to validate URLs in errorsHelp.json

kennethkalmer:
As part of reviewing #62 I realised we can automate this and at any point in the future easily validate the destination URLs we send users to.

This has been implemented with Ruby & Rake. Running `rake` in the `protocol` folder will validate each link in `errorsHelp.json` by ensuring they return a 200 response code.

If needed it can easily be rewritten in another in another language.


lmars (approved):
LGTM

SimonWoolf (approved):

kennethkalmer:
@QuintinWillison pointed out that this repo is required by many of our other repos via git submodules, I didn't realise. Would it be a problem to ship these extra files along to those repos? I personally don't think so, just would love to get others take on this.

lmars:
Not a problem in my opinion

PR #65, opened 2021-02-17
new control api limits: new error codes

rmbellovin:


protocol/errors.json
```diff
@@ -43,7 +43,11 @@
 	"40112": "account blocked (message limits exceeded)",
 	"40113": "account blocked",
 	"40114": "account restricted (channel limits exceeded)",
+	"40115": "account restricted (application limits exceeded)",
```
SimonWoolf, line 46: > "account restricted (application limits exceeded)"

'application limits' might be a bit ambiguous - not quite clear that it means 'max number of applications' vs 'Some limit at the level of the application'

Alos an account restriction is a specific state that application can go into and out of, which isn't used for app limits. The distinction is a bit subtle, but still.

I'd suggest "maximum permitted number of applications exceeded", "maximum permitted number of rules exceeded", etc

SimonWoolf (approved):

PR #67, opened 2021-04-07
Update errorsHelp.json with additional error codes and links

markcoxsmith:
New article codes and links added: 40100, 40102, 40114, 50001, 70002, 80002, 80003, 80019, 80021, 90021, 91001, 91005


rmbellovin (approved):
LGTM

PR #68, opened 2021-06-15
Add initial agents.json

owenpearson:


protocol/agents.json
```diff
@@ -0,0 +1,85 @@
...
+            ]
+        },
+        {
+            "identifier": "ably-ruby-rest",
+            "versioned": false,
+            "emitters": [
+                "ably-ruby"
+            ]
+        }
+    ]
```
lmars, line 83: Should we include the current agents being sent in `X-Ably-Lib` (e.g. `ably-go`) so we can maintain some continuity in the stats and fix the current issues were having with `X-Ably-Lib` parsing?

protocol/agents.json
```diff
@@ -0,0 +1,85 @@
...
+            ]
+        },
+        {
+            "identifier": "ably-cocoa",
+            "emitters": [
+                "ably-cocoa"
+            ]
+        },
+        {
+            "identifier": "iOS",
```
lmars, line 29: Two of the product requirements are:

> Usage of different wrapper libraries (flutter, asset-tracking, etc)
> What versions of OSs and runtime environments are used?

So I think it would be good to add some extra classification of each agent in this list (e.g. `agent_type: "sdk|os|wrapper"`) that will be emitted along with the metric, which will simplify getting answers to the above questions with queries like:

- `SELECT * FROM agents WHERE agent_type = 'os'`
- `SELECT * FROM agents WHERE agent_type = 'wrapper'`

protocol/agents.json
owenpearson: I agree in principle, although I can't see an obvious way to do this. We'd presumable need to include the name of the `X-Ably-Lib` identifier which will be different than the equivalent `Ably-Agent` identifier and may optionally include a wrapper library. For example it would need to parse `jsrn-1.2.9` as `ably-js/1.2.9 react-native`.

protocol/agents.json
QuintinWillison: I like that idea, but maybe the root `agents` value could be a map where the keys of that map are the agent type and then the values are arrays. Given that, if I interpret correctly, an agent will only ever have one type.

QuintinWillison (commented):
What would be really nice would be a workflow to check this file has expected layout. Akin to what you're already doing in [ably-asset-tracking-common](https://github.com/ably/ably-asset-tracking-common) with the [JSON schema validator](https://www.npmjs.com/package/jsonschema).

protocol/agents.json
owenpearson: For javascript I imagine something like this could work:
```
        {
            "identifier": "ably-js",
            "emitters": [
                "ably-js"
            ],
            "x-ably-lib": {
                "identifier": "js",
                "variants": ["web", "node", "rn", "ns"]
            }
        },
```
This at least describes to the router exactly what the `X-Ably-Lib` is made up of. Actually using this might be more complicated - for example, you would need to account for the fact that `X-Ably-Lib` headers would simply indicate whether NodeJS is being used while `Ably-Agent` will always give you the NodeJS version if present.

protocol/agents.json
owenpearson: Either of these options sound good to me, @lmars @cruickshankpg do either of you have any preference so I can go ahead and add this?

protocol/agents.json
lmars: Hmm ok yes, I didn't realise the X-Ably-Lib version was so ad-hoc.

Looking a bit closer at the current router behaviour, it's even more broken than I first realised since it [splits on the first hyphen](https://github.com/ably/go-services/blob/main/router/request_handler.go#L553), meaning something like `js-web-1.2.9` is currently being emitted as `agent=js version=web-1.2.9` which definitely wasn't the intention.

So, what I'd suggest is defining an explicit mapping from X-Ably-Lib values to their equivalent Ably-Agent value, and use some convention to indicate "no version specified" for Ably-Agent entries which would have a version if provided directly but is missing from the legacy X-Ably-Lib values (e.g. `js-node-1.2.9` maps to `ably-js/1.2.9 nodejs/0.0.0`).

For example, in `agents.json`, have a section like this:

```json
"x-ably-lib-mappings": {
  "js-web":  "ably-js/${version} browser",
  "js-node": "ably-js/${version} nodejs/0.0.0",
  "js-ns":   "ably-js/${version} nativescript",
  "js-rn":   "ably-js/${version} reactnative",
  "go":      "ably-go/${version}"
}
```

and the router would take any `X-Ably-Lib` value that matches `<key>-x.x.x`, lookup `<key>` in the mappings, replace `${version}` with `x.x.x` in the mapped value, then process as if it were an Ably-Agent value.

This makes the handling of X-Ably-Lib pretty clear from just looking at those mappings in `agents.json`, and keeps the router logic pretty generic.

WDYT?

cc @cruickshankpg

protocol/agents.json
lmars: From a router perspective, this would be easier as fields on each entry so it can be deserialised into a struct like:

```go
type Agent struct {
	Identifier string `json:"identifier"`
	Type       string `json:"agent_type"`
}
```

then when we do the lookup to find the `Agent` object for a given string, we can easily just reference `agent.Type` when emitting the metric.

protocol/agents.json
QuintinWillison: Related comment: let's not use `kebab-case` for key names - it's only necessary for literal values.

protocol/agents.json
owenpearson: I've added a type property in 8896442

protocol/agents.json
owenpearson: added the "x-ably-lib-mappings" property in b7a883e

owenpearson: and changed it to camelCase in b7a883e

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "ably-common",
+  "version": "1.0.0",
+  "scripts": {
+    "test": "jest"
+  },
+  "repository": {
```
QuintinWillison, line 7: Because we're using GitHub this can be shortened to just:

```json
"repository": "ably/ably-common"
```

As documented [here](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository) and demonstrated by us [here](https://github.com/ably/ably-js/blob/b1ef7473d5b8ca9a3a94de543f547ccfd6e6e71d/package.json#L79).

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "ably-common",
+  "version": "1.0.0",
```
QuintinWillison, line 3: Perhaps start this off at `1.2.0` to represent protocol version as it stands right now?

test-resources/agents-schema.json
```diff
@@ -0,0 +1,49 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Ably Agents Definition",
+    "id": "https://schemas.ably.com/json/asset-common/Agetns",
```
QuintinWillison, line 4:
```suggestion
    "id": "https://schemas.ably.com/json/asset-common/Agents",
```

QuintinWillison (commented):
Also, I notice that your schema explicitly sets `additionalProperties` in a couple of places to `true`... ignoring the fact that `true` is, as far as I can see, the default value 😉 - I think it would be wiser to nail this schema down really tight from the outset. If new properties need adding in future then we should adjust the schema atomically with the JSON at that time.

package.json
owenpearson: Good idea aee5762

package.json
owenpearson: Nice, didn't know about this 2609d56

QuintinWillison (approved):

.github/workflows/check.yml
```diff
@@ -12,6 +12,8 @@ jobs:
 
       # Use the jq tool (included in GitHub runner environments) to simplistically validate JSON files.
       - run: |
+          npm ci
           cd protocol
           jq empty errors.json
           jq empty errorsHelp.json
+          npm t
```
lmars, line 19: I'm a firm believer that commands in scripts should always use the longest (most readable) form, so `npm test` rather than `npm t` (assuming that's what is being run here)

protocol/agents.json
```diff
@@ -0,0 +1,118 @@
...
+                "ably-js"
+            ],
+            "type": "sdk"
+        },
+        {
+            "identifier": "node",
+            "emitters": [
+                "ably-js"
+            ]
+        },
```
lmars, line 15: For consistency, can we give all entries a `type`, even if it's `type: "misc"` or something? Specifically here, I'd probably expect `node` to be `type: "runtime"`

protocol/agents.json
```diff
@@ -0,0 +1,118 @@
...
+            "identifier": "ably-ruby-rest",
+            "versioned": false,
+            "emitters": [
+                "ably-ruby"
+            ],
+            "type": "wrapper"
+        }
+    ],
+    "ablyLibMappings": {
+        "js-web": "ably-js/${version} browser",
```
lmars, line 96: All the agent values that appear here should be in the `agents` array too (i.e. `browser`, `nativescript`, `uwp` etc)

test-resources/agents-schema.json
```diff
@@ -0,0 +1,47 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Ably Agents Definition",
+    "id": "https://schemas.ably.com/json/asset-common/Agents",
```
lmars, line 4: Should we set up `schemas.ably.com`? We're also going to have JSON schemas for plugins (see [here](https://ably.atlassian.net/wiki/spaces/ENG/pages/1197342737/Ably+Plugins+Architecture#Parameters)), and so far we've referenced the schemas under the fictitious `plugins.ably.com` domain, but would be good to standardise this (separate from this PR, of course).

test-resources/agents-schema.json
```diff
@@ -0,0 +1,47 @@
...
+                        "type": "string",
+                        "enum": ["os", "sdk", "wrapper"]
+                    }
+                },
+                "required": ["identifier", "emitters"]
+            }
+        },
+        "ablyLibMappings": {
+            "type": "object",
+            "patternProperties": {
```
lmars, line 38: Let's give this a description

protocol/agents.json
QuintinWillison: Is there a reason why it can't be optional? i.e. if it's not there (`undefined`) then it's implicitly "misc". 🤔

test-resources/agents-schema.json
QuintinWillison: This isn't the first time this has come up. I'm trying to find previous conversation(s)...

QuintinWillison: Ah, yes: https://github.com/ably/ably-asset-tracking-common/blob/170429211965e098e211bc117d29eeceef405e64/test-resources/geo/journey-schema.json#L9

test-resources/agents-schema.json
owenpearson: done ddacc85

test-resources/agents-schema.json
QuintinWillison: Ideally I think we would want to do https://github.com/ably/ably-common/issues/70 first and then have a formal release procedure, which could also publish schemas out via S3 or whatever. (apologies for monologue 😆)

protocol/agents.json
lmars: I just feel like if you have an array of typed things, they should each have a type, and I'm a proponent of explicitness, especially in "common" things like this which is going to be shared between a variety of systems and users (SDKs, router, Grafana dashboards, Athena queries)

test-resources/agents-schema.json
owenpearson: I'm happy for us to set up the subdomain and host these at some point but personally I don't see the problem with them being fictitious either. I do see the value in standardising this though, so I would be in favour of changing the plugin schemas to use `schemas.ably.com`.

lmars: Sounds good, perhaps we'll want a `isSchema` option (or equivalent) for the sdk-upload-action so repositories can publish their own schemas to `schemas.ably.com` using GitHub actions?

protocol/agents.json
owenpearson: Either way, all the current identifiers fit nicely into the `sdk|os|wrapper|runtime` tetrachotomy so no need for misc at the moment. fixed in e41c688.

.github/workflows/check.yml
owenpearson: agreed, this was just force of habit on my part 06a6694

protocol/agents.json
owenpearson: done 23e8f2c

protocol/agents.json
```diff
@@ -0,0 +1,195 @@
...
+        "java": "ably-java/${version}",
+        "cocoa.ios": "ably-cocoa/${version} iOS/0.0.0",
+        "cocoa.macos": "ably-cocoa/${version} macOS/0.0.0",
+        "cocoa.tvos": "ably-cocoa/${version} tvOS/0.0.0",
+        "dotnet": "ably-dotnet/${version}",
+        "dotnet.framework": "ably-dotnet/${version} dotnet-framework/0.0.0",
+        "dotnet.uwp": "ably-dotnet/${version} uwp",
+        "dotnet.netstandard20": "ably-dotnet/${version} dotnet-standard/0.0.0",
+        "dotnet.xamarin-android": "ably-dotnet/${version} xamarin android/0.0.0",
+        "dotnet.xamarin-ios": "ably-dotnet/${version} xamarin iOS/0.0.0",
```
tomkirbygreen, line 188: I think we might need a "dotnet.unity" here. Perhaps something like:

```
"dotnet.xamarin-ios": "ably-dotnet/${version} unity/0.0.0",
```

tomkirbygreen (commented):
Just a point regarding recognising Unity as a supported .NET platform / target.

protocol/agents.json
owenpearson: I've added unity as a new identifier in 5b317f5

lmars (approved):

tomkirbygreen (commented):
Looks good to me.

test/agents.test.js
```diff
@@ -0,0 +1,21 @@
...
+const schema = require(schemaPath);
+const agents = require(agentsPath);
+
+const jsonschema = new Validator();
+jsonschema.addSchema(
+    schema,
+    'https://schemas.ably.com/json/asset-common/Agents'
+);
+
+describe("agents.json", () => {
```
mattheworiordan, line 16: I recognise I am being pedantic, but in this file we're mixing single quotes and double quotes for strings.

test-resources/agents-schema.json
```diff
@@ -0,0 +1,49 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
```
mattheworiordan, line 2: This is awesome, I love the fact you've set up a schema for this

test/agents.test.js
owenpearson: Ah, my bad. I'll make sure to get prettier running on JS files when we resolve #79 which will fix this along with any other styling inconsistencies.

PR #74, opened 2021-06-17
Conform protocol JSON files

QuintinWillison:
Make the JSON legal, for starters. 😁 

There's more we can do in due course, but this is just a quick improvement to give us a more maintainable repo..

### DIFF
--- a/protocol/README.md
+++ b/protocol/README.md
@@ -1,7 +1,36 @@
-## Ably Realtime Error Codes
+# Ably Realtime Error Codes
+
+## Overview
 
 Ably's [client library SDKs](https://www.ably.io/download), and the [core realtime platform](https://www.ably.io/platform), use common error codes. These error codes are stored and maintained in the shared public JSON file [errors.json](./errors.json).
 
 In addition, a list of error codes and corresponding support articles to help understand how to resolve common problems relating to these error codes can be seen in [errorsHelp.json](./errorsHelp.json).
 
 If you need help understanding any error codes, or need technical support, please visit the [Ably support desk](https://www.ably.io/support).
+
+## Error Code Ranges
+
+Codes in [error.json](error.json) fall into these ranges:
+
+| From  | Title |
+| ----- | ----- |
+| 10000 | Generic |
+| 40000 | 400 |
+| 40100 | 401 |
+| 40300 | 403 |
+| 40400 | 404 |
+| 40500 | 405 |
+| 41001 | 410 |
+| 42200 | 422 |
+| 42910 | 429 |
+| 50000 | 500 |
+| 50010 | Edge cacheing / proxy service error codes |
+| 50320 | Active Traffic Management error code to indicate intentional redirect of traffic to fallback hosts |
+| 60000 | reserved for internal (non-customer-facing) use |
+| 70000 | reactor-related |
+| 71000 | exchange-related: general |
+| 71100 | exchange-related: publisher |
+| 71200 | exchange-related: product |
+| 71300 | exchange-related: subscription |
+| 80000 | connection-related |
+| 90000 | channel-related |
--- a/protocol/README.md
+++ b/protocol/README.md
@@ -24,8 +24,10 @@ Codes in [error.json](error.json) fall into these ranges:
 | 42200 | 422 |
 | 42910 | 429 |
 | 50000 | 500 |
-| 50010 | Edge cacheing / proxy service error codes |
+| 50010 | Edge cacheing / proxy service |
 | 50320 | Active Traffic Management error code to indicate intentional redirect of traffic to fallback hosts |
+| 50330 | DNS switch over |
+| 50410 | Edge cacheing / proxy service |
 | 60000 | reserved for internal (non-customer-facing) use |
 | 70000 | reactor-related |
 | 71000 | exchange-related: general |
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -88,7 +88,7 @@
 	"50002": "internal connection error",
 	"50003": "timeout error",
 	"50004": "Request failed due to overloaded instance",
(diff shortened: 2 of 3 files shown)


QuintinWillison:
[Oh dear](https://github.com/ably/ably-common/actions/runs/946789948) 🙄 ... looking...

> You have an error in your yaml syntax on line 16

QuintinWillison:
Despite my inability to do yaml, this is definitely ready for review now. Sorry for the noise.

owenpearson (approved):

tomkirbygreen (approved):

protocol/errors.json
```diff
@@ -73,71 +69,54 @@
...
 	"42921": "max per-connection publish rate limit exceeded (fatal); closing connection",
 
-	/* 500 codes */
 	"50000": "internal error",
 	"50001": "internal channel error",
 	"50002": "internal connection error",
 	"50003": "timeout error",
 	"50004": "Request failed due to overloaded instance",
-	"50330": "request reached the wrong cluster; retry (used during dns changes for cluster migrations)",
+	"50005": "request reached the wrong cluster; retry (used during dns changes for cluster migrations)",
```
lmars, line 91: Why has this changed? Searching for `50005` in the realtime code base, it's emitted [when the service is in lockdown](https://github.com/ably/realtime/blob/main/frontend/lib/service/rest.js#L37), which doesn't seem to match this description here.

QuintinWillison: Oh, @SimonWoolf and I discussed over Slack and he thought it would be ok to rename as it's not always in use. See my additional detail on this commit's message: https://github.com/ably/ably-common/pull/74/commits/1d262b37386a8bc2a630011258d4a560d48eee02

lmars: Hmm but I think we do set the code to 503 when we do the DNS switch over and I think that's part of the playbooks, just seemed odd to re-number it here amongst other changes. Anyway, If I see the code in a doc somewhere in the future I'll try and remember to change it 👍

SimonWoolf: Sorry, yes, as Lewis says 50005 is already used by something else, but it never got added to ably-common, which is my fault. I'll do a pr to add it, but if you're renaming this then need to pick something else - I don't think 50006 is in use if you want something in the 500xx block

lmars: Or just keep the code for this description the same and move it down in the file to after `50320`?

protocol/README.md
```diff
@@ -1,7 +1,36 @@
...
+| 40100 | 401 |
+| 40300 | 403 |
+| 40400 | 404 |
+| 40500 | 405 |
+| 41001 | 410 |
+| 42200 | 422 |
+| 42910 | 429 |
+| 50000 | 500 |
+| 50010 | Edge cacheing / proxy service error codes |
+| 50320 | Active Traffic Management error code to indicate intentional redirect of traffic to fallback hosts |
```
SimonWoolf, line 28: Not sure that this categorization really fits how the codes are used right now. E.g. 'edge caching' doesn't consume the whole block of three hundred codes from 50010 to 50310 - rather edge caching codes are a few normal http status codes with "10" added to the end. The router uses a similar thing but with '05', so 50205/50305/50405. The idea's was to have ties between the codes and the http status codes - have them be status codes with 'extra detail'. (though we don't do that consistently).

(I realise this isn't a new problem with this pr, the json comments gave a similar impression)

No strong opinion on whether this is actually a problem and if so what the solution is, happy for Quintin to say whether/how he'd like to redesign any of this, whether in the direction of stronger ties to http status codes, or weaker, or whatever

protocol/errors.json
SimonWoolf: cf comment https://github.com/ably/ably-common/pull/74#discussion_r653714585 - seems like the root of the problem is we currently have a bit of both of 'have functionally-related blocks of adjacent codes' and 'have codes be strongly tied to http status codes', and those are sometimes in tension

SimonWoolf (approved):

protocol/errors.json
QuintinWillison: Yes. @lmars, it may not be immediately obvious but I was also trying to gain some hold of the ranges / groups. See https://github.com/ably/ably-common/blob/feature/conform-protocol-json/protocol/README.md#error-code-ranges

protocol/README.md
QuintinWillison: I considered adding a "To" column but I would have been plucking those out of thin air. Hence I went with this lighter weight approach, directly reflecting what was in the existing JSON comments, albeit that only works if they're in order!

Perhaps the next iteration, ideally beyond the scope of this PR, would be to add that To column.

protocol/errors.json
QuintinWillison: On balance, I think I'll add a commit to revert this code back to where it was, and as @lmars suggests position it correctly so everything's in sequence. I think it's going to also get its own range too - I'll also include that in the commit.

QuintinWillison: https://github.com/ably/ably-common/pull/74/commits/e671455fbbcfa1b00821332c909f563a83808be6

@lmars please could you resolve this conversation if that fixes this for you?

protocol/README.md
QuintinWillison: https://github.com/ably/ably-common/issues/76 👈  @SimonWoolf, if you're happy with that deferment then could you please resolve this conversation?

protocol/errors.json
lmars: 👍

lmars (approved):

protocol/README.md
QuintinWillison: I'm going to resolve this conversation on your behalf, @SimonWoolf, as a prerequisite to landing the PR. If you need to follow this up further then please do so under #76. Thanks.

PR #78, opened 2021-06-22
Conform whitespace

QuintinWillison:
Gets rid of the mix of tab and space characters in source files (including within the same file at times 🤯).

Also adds a trivial, high-level check to ensure they don't slip in again in future now that we have CI running.


tomkirbygreen (approved):
Looks good to me ('Spaces for the win!')

owenpearson (approved):
LGTM, but I would prefer using the NPM prettier module for checking in CI since it can check all the provided rules, not just tabs vs spaces (and we can also add an NPM script to format all the files). Happy to leave that for a separate PR though, I can have a look at doing this later on today.

.github/workflows/check.yml
```diff
@@ -10,8 +10,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # Use the jq tool (included in GitHub runner environments) to simplistically validate JSON files.
+      # Using git grep to fail the build if any tab characters are found in source files - we use spaces for whitespace.
+      # Using the jq tool (included in GitHub runner environments) to simplistically validate JSON files.
       - run: |
+          ! git --no-pager grep $'\t'
```
lmars, line 16: I wonder if it's worth outputting something like:

"Lint failed! Source files must use spaces for whitespace, but the following files use tabs: `<insert list>`"

Otherwise I feel the build failure would be pretty cryptic?

lmars (approved):

.github/workflows/check.yml
QuintinWillison: Hopefully this addresses your concern (albeit not exactly the same form you suggest): https://github.com/ably/ably-common/pull/78/commits/3e48a5664917cc9b5d57453c2b5db6b0fd7b1a6e

QuintinWillison:
@owenpearson I've addressed [your comment](https://github.com/ably/ably-common/pull/78#pullrequestreview-689251314) by creating #79.

Though I'll admit that I'm not so keen on the idea of having an NPM script to format files as I believe that should be a natural part of the editing workflow, rather than an "after thought" - hence my addition of `.editorconfig`.

.github/workflows/check.yml
lmars: That'll do 🙂

PR #82, opened 2021-06-22
Make versioned field explicit in agents.json

lmars:
This prevents code that naively parses the versioned field into a bool treating an unspecified value as false when the intention was for an unspecified value to be true.


owenpearson (approved):
LGTM

Issue #70, opened 2021-06-16
Add Version


QuintinWillison:
As we do for our other code repositories, this repository should have:

* a version number
* a "release" procedure
* a change log

The version is going to effectively end up closely, or exactly, representing the Ably protocol version.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-2073) by [Unito](https://www.unito.io)

QuintinWillison:
Perhaps under this issue, or perhaps under a new issue, we should also consider publishing some of the artifacts contained (or generated) here out to package repositories, in order to remove the need to use Git [submodules](https://git-scm.com/docs/gitsubmodules) downstream. 🤔

QuintinWillison:
Work under this issue should include, or at least generate, work on [ably-common-go](https://github.com/ably/ably-common-go) so that the version is propagated to there.

See [this internal pull request](https://github.com/ably/go-services/pull/600) for context.

QuintinWillison:
This work is anticipated to be done by the SDK Team, probably myself, so I am adding the `sdk` label to indicate that.

lmars:
@QuintinWillison apologies for not replying sooner. I question this statement in the description:

> The version is going to effectively end up closely, or exactly, representing the Ably protocol version.

Could you perhaps expand on what about the contents of ably-common you think implies that the version should either closely or exactly follow the Ably protocol version?

QuintinWillison:
We've had valuable discussion around this under [this internal Slack thread](https://ably-real-time.slack.com/archives/C030C5YLY/p1657269348759789) started by @Morganamilo.

Within that discussion I stated:

> I've had a niggling feeling, for some time, that I've shoehorned the features work into `ably-common`. Would anybody object to an assertive move / change of tack, whereby we create a brand new repository in the `ably` org that is simply called `features` (yes, not even a need for the superfluous, org-context-ignoring `ably-` prefix :wink:). That is then the new home for the current contents of the `features/` folder from `ably-common`.
> At that point, protocol is - indeed - kept discrete, as its own concern.
> And, I hope, nobody can object to that new repository being semantically versioned.

To which I added, in relation to what's on the roadmap after working on this ticket:

[...]

PR #87, opened 2021-06-24
Add auto-generated Go code from protocol/agents.json

lmars:
This introduces a `go` directory with Go code generated from `protocol/agents.json` to make the list of agents easier to reference from internal services.

I've also added `scripts/generate.sh` which generates the Go code, and it includes a `--check` flag which is run in CI to make sure generated files are up-to-date.


paddybyers:
Elsewhere we generate and check via makefile targets. Should we do that here?

lmars:
> Elsewhere we generate and check via makefile targets. Should we do that here?

I defer that to @QuintinWillison.

cruickshankpg (approved):
I'm slightly nervous that generate.sh may grow into a giant unmanageable bash file if more languages are added but I suppose it's a reasonable solution for now. Otherwise LGTM

QuintinWillison (changes requested):
I'm not happy with committing generated files back to this repository. I would prefer that this repository is purely _source_ files. I'm going to work with @lmars to explore alternatives.

lmars:
@QuintinWillison I've been thinking about the desire to have the generated code somewhere separate that's updated automatically by a GitHub action, and there are two options I can think of:

1. Upload the generated code to the `sdk.ably.com` S3 bucket
2. Commit the generated code to a separate repository (e.g. github.com/ably/ably-common-generated)

S3 is simpler from a workflow perspective since we already have that pipeline setup for our SDK repos, but it's not trivial for Go to pull code directly from S3, it expects modules to be located in a version control system.

A separate repository is simpler from a Go perspective as it just looks like a normal module dependency, but having a GitHub action pushing changes to that repository seems fiddly.

For both options though, it will require our internal Go services to reference two different modules: ably-common for the definition of the types (`Agent`, `Entry`, etc.), and S3/ably-common-generated for the generated lists. Also, the generated code will need to depend back on ably-common to reference the types in those lists.

This is quite a bit of added complexity to avoid having generated code in this repository, so I think we should either:

- make an exception for Go, allow generated code to be committed to this repository, and reference it as a single versioned module from our internal repository
- revert back to the original implementation where the types and generated code lived within our internal repository with ably-common referenced as a submodule

I'd much prefer the former to the latter, but I will let you decide and then act accordingly.

QuintinWillison:
@lmars Shame the S3 route wouldn't be a slam dunk, so let's ignore that.

In terms of the downstream repository idea, especially in light of planned #70, I feel that a formal (manually triggered) versioned published from this repository (typically `main` branch, after a tag push) is the right way to go forwards. Is there a reason why that publish couldn't include as published artifacts the other files you need from this repository? - i.e. a snapshot containing "raw" and "generated" files. In which case I might propose we specifically call that downstream repository `ably-common-go` to represent the fact that it's designed to be consumed according to the specific needs of Golang.

I would propose modelling that publish workflow on [what we did in asset tracking](https://github.com/ably/ably-asset-tracking-android/blob/main/.github/workflows/publish.yml) (just commit and push to downstream Git repo, rather than [packages](https://docs.github.com/en/packages)).

cruickshankpg:
> For both options though, it will require our internal Go services to reference two different modules: ably-common for the definition of the types (Agent, Entry, etc.), and S3/ably-common-generated for the generated lists. Also, the generated code will need to depend back on ably-common to reference the types in those lists.

if you also copy across the type definitions to the generated repo then that shouldn't be needed

lmars:
> I would propose modelling that publish workflow on what we did in asset tracking (just commit and push to downstream Git repo, rather than packages).

Thanks. I took a look at the asset tracking workflow and it looks like it calls [`./gradlew publish`](https://github.com/ably/ably-asset-tracking-android/blob/main/.github/workflows/publish.yml#L30) which I assume runs [this code](https://github.com/ably/ably-asset-tracking-android/blob/main/publish.gradle), isn't that just publishing to Maven though?

> If you also copy across the type definitions to the generated repo then that shouldn't be needed

Good point 👍

lmars:
A reply from @QuintinWillison from an internal discussion:

> Oh, sorry, we don't do this yet. You're right, what I pointed you towards was Maven via GitHub packages. For Git we will probably need to use appropriate GitHub-provided actions I would suggest. The workflow I was pointing you t'wards was more around how I manually trigger it and checkout a tag.

So I'm going to put something together to do the git publish.

lmars:
@QuintinWillison I've just pushed f531984 which adds a manually triggered workflow to publish generated Go code to the [ably-common-go](https://github.com/ably/ably-common-go) repository.

I created a deploy key on the ably-common-go repository, set the deploy key as the `ABLY_COMMON_GO_SSH_KEY` secret on this repository, and the publish workflow uses that with the [webfactory/ssh-agent](https://github.com/webfactory/ssh-agent) action to gain write permissions to ably-common-go.

I've successfully run the script locally (and then force pushed ably-common-go main back), but I've so far been unable to trigger the workflow as I guess it needs to be merged first?

PR #93, opened 2021-07-13
Expand agents documentation about python rest agent

d8x:
https://github.com/ably/ably-python/issues/168
https://github.com/ably/docs/pull/1034
- Add python rest agent information

### DIFF
--- a/protocol/agents.json
+++ b/protocol/agents.json
@@ -171,20 +171,20 @@
             "type": "runtime"
         },
         {
+          "identifier": "python",
+          "versioned": true,
+          "emitters": [
+            "ably-python"
+          ],
+          "type": "runtime"
+        },
+        {
             "identifier": "ably-python",
             "versioned": true,
             "emitters": [
                 "ably-python"
             ],
             "type": "sdk"
-        },
-        {
-          "identifier": "ably-python-rest",
-          "versioned": true,
-          "emitters": [
-              "ably-python"
-          ],
-          "type": "wrapper"
         },
         {
             "identifier": "ably-php",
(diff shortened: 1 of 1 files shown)


protocol/agents.json
```diff
@@ -223,6 +231,7 @@
         "php": "ably-php/${version}",
         "php-laravel": "ably-php/${version} laravel",
         "python": "ably-python/${version}",
+        "python-rest": "ably-python/${version} ably-python-rest",
```
lmars, line 234: The entries in `ablyLibMappings` are used to map deprecated `X-Ably-Lib` headers to their `Ably-Agent` equivalent, so given ably-python has never set `python-rest` in the `X-Ably-Lib` header, this mapping is not needed.

protocol/agents.json
```diff
@@ -178,6 +178,14 @@
             ],
             "type": "sdk"
         },
+        {
+          "identifier": "ably-python-rest",
+          "versioned": true,
```
lmars, line 183: https://github.com/ably/ably-python/pull/195 introduces `ably-python-rest` as an entry without a version, so this here should be `versioned: false`

lmars (commented):
This PR should also add a versioned agent entry for `python`

protocol/agents.json
lmars: Then again, I don't actually think `ably-python-rest` is required, since I believe there is only one ably-python package (as opposed to Ruby where you can either install [ably-ruby](https://github.com/ably/ably-ruby) or [ably-ruby-rest](https://github.com/ably/ably-ruby-rest)).

d8x: https://github.com/ably/ably-common/pull/93/commits/7483964788a44ca91389519a129e430d10925d28

d8x: https://github.com/ably/ably-python/pull/195/commits/a5c237791504cfbc9e47e57c1e408987d03f84a4

lmars: nice 👍

lmars (approved):
:shipit:

QuintinWillison (approved):

PR #96, opened 2021-07-26
Expand commentary around demo repositories

QuintinWillison:
We had some interesting conversation in our internal company meeting on Friday around demos and Ably use case examples.

What became apparent is that there are some valuable demos coming out of our developer relations team which need to be made more visible to those visiting our client library SDK repositories. This pull request attempts to make that a bit clearer.


marklewin (approved):
LGTM

ben-xD (approved):

PR #98, opened 2021-07-26
Added go runtime identifier

sacOO7:


protocol/agents.json
```diff
@@ -234,6 +234,14 @@
                 "ably-go"
             ],
             "type": "sdk"
+        },
+        {
+            "identifier": "go",
+            "versioned": true,
+            "emitters": [
+              "ably-go"
```
lmars, line 242: Please use a 4-space indent

sacOO7: I don't have indentation tool : (
Seems need to do it manually

sacOO7: Fixed !

lmars (approved):
Thanks!

PR #99, opened 2021-07-30
Updated Go agent

sacOO7:
- Added ably-go as emitter for darwin. windows and linux


protocol/agents.json
```diff
@@ -82,19 +82,29 @@
             ],
             "type": "os"
         },
+        {
+            "identifier": "darwin",
```
tomkirbygreen, line 86: How does Darwin differ from macOS? Surely they're one and the same< which is to say I don't think we need to add Darwin.

tomkirbygreen (changes requested):

protocol/agents.json
sacOO7: Updated to `macOS-iOS`

QuintinWillison: Let's stick with `darwin`. I think it may still be of use as from what I can tell we can probably infer macOS version from Darwin version when we look at logs later on (see [here](https://en.wikipedia.org/wiki/Darwin_(operating_system)#Release_history)).

sacOO7: Done !

QuintinWillison (approved):

tomkirbygreen (approved):

PR #111, opened 2021-09-09
Conform formatting of JSON files

QuintinWillison:
Fixes #79.

PR #112, opened 2021-09-09
Contributing Guide

QuintinWillison:
There's probably more we could add in due course, but I think this is a good starting point.

My focus, for the moment, was mainly relating to the changes made in [the preceding pull request](https://github.com/ably/ably-common/pull/111).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,29 @@
+# Contributing to Ably's Common Resources
+
+## Development
+
+### Protocol
+
+See [Ably Realtime Protocol](protocol) for details on maintaining:
+
+- Error Codes
+- Error Help (support article links)
+- Agents
+
+### JSON Files
+
+Our [CI check workflow](.github/workflows/check.yml) runs [Prettier](https://prettier.io/) over all JSON files in this repository.
+
+If you want to be sure that changes you make to JSON files will not fail CI then you can either use Prettier's `--check` switch as implemented in the aforementioned check workflow, or you can simply ask Prettier to rewrite incorrectly formatted JSON files with this command:
+
+    npx prettier --write "**/*.json"
+
+## Release Process
+
+The outputs from this repository are not yet versioned, though this is planned ([#70](https://github.com/ably/ably-common/issues/70)).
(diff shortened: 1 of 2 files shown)


tomkirbygreen (approved):

owenpearson (approved):

marklewin (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Error Codes
+- Error Help (support article links)
+- Agents
+
+### JSON Files
+
+Our [CI check workflow](.github/workflows/check.yml) runs [Prettier](https://prettier.io/) over all JSON files in this repository.
+
+If you want to be sure that changes you make to JSON files will not fail CI then you can either use Prettier's `--check` switch as implemented in the aforementioned check workflow, or you can simply ask Prettier to rewrite incorrectly formatted JSON files with this command:
```
lmars, line 17: Is it perhaps worth stating explicitly that JSON files should use two space indent? That seems to be the biggest source of formatting differences between contributions so far.

lmars (approved):

ben-xD (approved):

CONTRIBUTING.md
QuintinWillison: For me it's enough to refer to [Prettier](https://prettier.io/) here, as we're using their defaults, and indentation is already specified in their [implementation](https://github.com/prettier/prettier/blob/24d39a906834cf449304dc684b280a5ca9a0a6d7/src/main/core-options.js#L236) and [docs](https://prettier.io/docs/en/options.html#tab-width). Also, we have [an EditorConfig config file](https://github.com/ably/ably-common/blob/main/.editorconfig) in this repository for those using an IDE aware of that standard (e.g. [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)). Explicitly documenting that here is repeating information already canonically expressed elsewhere (it could fall out-of-sync, albeit unlikely).

lmars: Fair enough 👍

PR #114, opened 2021-09-18
Add GitHub labels

### LINKED ISSUE #108: Consider YAML as source format for schemas and other JSON documents
I found [this comment](https://github.com/json-schema-org/json-schema-spec/issues/100#issuecomment-254365011) and it really got me thinking.

My context at the time was wondering how I could be more expressive, in a human readable form, in `description` fields in JSON schema. It was affirming, albeit unsurprising, to find that others feel the same way.

Perhaps we should also think of YAML as for humans and JSON as for machines. 🤔

QuintinWillison:
### Immediate Value Delivered

The primary focus of this pull request is to add `github/labels.yml` as an initial import for a canonical location for us to document the GitHub labels we're using across our open source SDK repositories.

The contents of this file will evolve over time, both in terms of adding new labels, modifying the ones that have been defined here, and perhaps even by adding new keys. The keys defined so far - being the name (top-level key, implicit), `description` and `color` - closely match what's defined in GitHub.

### Future Plan

A later development, potentially in this codebase but perhaps in a new codebase (yet to be defined) downstream of this one, is that we will use this file as source input to a script that uses GitHub's GraphQL REST API to validate labels defined across all of our repositories (as part of SDK monitoring / auditing / conforming initiative).

Much of this future is in my head right now (early ideation / R&D / prototyping stage) - so feel free to reach out to me directly if you would like more information on that.

### Why Add YAML?

The choice of YAML for this file, adding to the file formats utilised by this repository, is intentional and partially experimental:

- This is **source code**: For human maintainers of this file, YAML is a lighter weight format that should be easier to wrangle in most editors.
- This can be transformed for **downstream consumption** (packaging / distribution): There is tooling readily available to convert from YAML to JSON, for example.

Thus, fixes #108.

### DIFF
--- a/github/labels.yml
+++ b/github/labels.yml
@@ -1,4 +1,3 @@
----
 api:
   description: Something that will change or improve the public API.
   color: ec4b42
--- a/github/labels.yml
+++ b/github/labels.yml
@@ -1,0 +1,31 @@
+---
+api:
+  description: Something that will change or improve the public API.
+  color: ec4b42
+bug:
+  description: Something isn't working. It's clear that this does need to be fixed.
+  color: d73a4a
+clarification-required:
+  description: This issue needs more detail.
+  color: e3d79b
+code-quality:
+  description: Affects the developer experience when working in our codebase.
+  color: ff88cc
+consistency:
+  description: Relates to ecosystem-level compatibility across implementations.
+  color: f4c45c
+documentation:
+  description: Improvements or additions to public interface documentation (API reference or readme).
+  color: 0075ca
+enhancement:
+  description: New feature or request.
+  color: a2eeef
+example-app:
+  description: Relates to the example apps included in this repository.
+  color: 70fc6b
+failing-test:
+  description: Where a test is failing either locally or in CI. Perhaps flakey (badly written), wrong or bug.
+  color: ff8888
+important:
+  description: We should try not to overuse this label, but it's a quick way to lightly mark as higher priority.
+  color: b60205
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -11,11 +11,12 @@ jobs:
       - uses: actions/checkout@v2
 
       # Using git grep to fail the build if any tab characters are found in source files (except Go files) - we use spaces for whitespace.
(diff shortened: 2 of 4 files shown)


QuintinWillison:
I've added quite a few reviewers. I certainly don't expect an approval from everybody from whom a review has been requested (though I would appreciate it!)... My reasons for adding this many reviewers were more around visibility as this change intentionally continues the journey we're already on in some places of 'documenting in code' so that we can automate later, even if we don't immediately.

github/labels.yml
```diff
@@ -0,0 +1,31 @@
+---
```
QuintinWillison, line 1: Potentially superfluous. 🤔

tomkirbygreen (approved):

lmars (approved):

paddybyers (approved):

ben-xD (approved):

github/labels.yml
QuintinWillison: https://github.com/ably/ably-common/pull/114/commits/7396dcfc15d751621cd9daced28f7a413f433370

jamienewcomb:
@QuintinWillison we currently have spec-alignment across our libraries but did not notice that here. What will we use in that ones place?

QuintinWillison:
Thanks, @jamienewcomb. That is a good point (e.g. as used [in ably-dotnet](https://github.com/ably/ably-dotnet/labels/spec-alignment)).

I was trying to keep the scope of this 'initial import' pull request as small as possible, so I focussed on bringing in the labels we had conformed to use in a more recently established repository.

We can follow up later with subsequent pull requests to add other labels upon which the team rely, including `spec-alignment`. For now I just want to get this file established, so that we can iterate it later. Until monitoring it has been automated, as described in my opening comment, it's low impact - and once monitoring has been established, gaps will be obvious (i.e. low risk).

SimonWoolf (approved):

AndyNicks (approved):

jamienewcomb (approved):

owenpearson (approved):

PR #117, opened 2021-09-27
Replacing third-party action with bash

QuintinWillison:
This came out of a conversation @lmars and myself had this morning. The third-party action could potentially leak the private key outside of the workflow runner.

At the time of writing this comment, this modification is not yet proven (we need to run this workflow manually).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,9 +15,12 @@ jobs:
           go-version: "1.16.5"
 
       - name: Setup SSH access
-        uses: webfactory/ssh-agent@v0.5.3
-        with:
-          ssh-private-key: ${{ secrets.ABLY_COMMON_GO_SSH_KEY }}
+        env:
+          SSH_PRIVATE_KEY: ${{ secrets.ABLY_COMMON_GO_SSH_KEY }}
+        run: |
+          echo "$SSH_PRIVATE_KEY" > ssh-private-key.pem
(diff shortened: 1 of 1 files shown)


owenpearson (approved):

.github/workflows/publish.yml
```diff
@@ -15,9 +15,12 @@ jobs:
...
       - name: Setup SSH access
-        uses: webfactory/ssh-agent@v0.5.3
-        with:
-          ssh-private-key: ${{ secrets.ABLY_COMMON_GO_SSH_KEY }}
+        env:
+          SSH_PRIVATE_KEY: ${{ secrets.ABLY_COMMON_GO_SSH_KEY }}
+        run: |
+          echo "$SSH_PRIVATE_KEY" > ssh-private-key.pem
+          eval $(ssh-agent -s)
+          ssh-add ssh-private-key.pem
```
lmars, line 23: Another thing that the webfactory/ssh-agent action does is add the GitHub keys to the known hosts file:

https://github.com/webfactory/ssh-agent/blob/master/index.js#L21-L22

This can also be achieved using `ssh-keyscan` (although obviously not pinned):

```
ssh-keyscan github.com > ~/.ssh/known_hosts
```

Although maybe GitHub actions already know about the GitHub host keys? 🤔 

Probably worth merging this, then if it breaks with `Host key verification failed`, follow up with a fix.

lmars (approved):

PR #118, opened 2021-09-27
[REA-980] Json-schemas: add metachannel client events

Jmgr:
This adds metachannel client events to the Json schemas.

### DIFF
--- a/json-schemas/src/client-events-api-requests.json
+++ b/json-schemas/src/client-events-api-requests.json
@@ -1,0 +1,41 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ably API Requests Client Event",
+  "type": "object",
+  "properties": {
+    "host": {
+      "type": "string",
+      "description": "the host the request was made to"
+    },
+    "requestId": {
+      "type": "string",
+      "description": "the unique ID of the request that Ably can use to correlate a request event with internal logs, if necessary"
+    },
+    "region": {
+      "type": "string",
+      "description": "the region the request was made in"
+    },
+    "headers": {
+      "type": "object",
+      "description": "the headers sent with the request"
+    },
+    "query": {
+      "type": "object",
+      "description": "the details of the parsed request query, including information such as the request format"
+    },
+    "path": {
+      "type": "string",
+      "description": "the path of the endpoint called in the request, for example /channels/{channel-name}/messages"
+    },
+    "channels": {
+      "type": "array",
+      "description": "a list of channels the request was made against"
+    },
+    "error": {
+      "type": "object",
+      "description": "the details of any error encountered whilst making the request. It includes an error message, error code and HTTP status code"
+    }
+  },
+  "additionalProperties": false,
+  "required": ["host", "region", "headers", "query", "path"]
+}
--- a/json-schemas/src/client-events-api-requests.json
+++ b/json-schemas/src/client-events-api-requests.json
@@ -18,23 +18,38 @@
     },
     "headers": {
       "type": "object",
-      "description": "The HTTP headers sent with the request. Keys should be lower case."
+      "description": "The HTTP headers sent with the request. Header names are lower case (e.g. 'accept', 'content-type', 'user-agent', etc.)."
     },
     "query": {
       "type": "object",
-      "description": "The details of the parsed request query, including information such as the request format."
+      "description": "The parsed query component of the API request URL (e.g. {\"format\":\"msgpack\"})."
     },
     "path": {
       "type": "string",
-      "description": "The path of the endpoint called in the request, for example /channels/{channel-name}/messages."
+      "description": "The path component of the API request URL (e.g. /channels/{channel-name}/messages)."
     },
     "channels": {
       "type": "array",
-      "description": "A list of channels the request was made against."
+      "description": "The list of channels the API request references."
     },
     "error": {
       "type": "object",
-      "description": "The details of any error encountered whilst making the request. It includes an error message, error code and HTTP status code."
+      "description": "The details of any error encountered whilst making the API request.",
+      "properties": {
+        "message": {
+          "type": "string",
+          "description": "The error message."
+        },
+        "code": {
+          "type": "integer",
+          "description": "The error code. See https://knowledge.ably.com/ably-error-codes."
+        },
+        "statusCode": {
+          "type": "integer",
+          "description": "The error status code. See https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html."
+        }
+      },
+      "required": ["message", "code", "statusCode"]
     }
   },
   "additionalProperties": false,
--- a/json-schemas/src/client-events-api-requests.json
+++ b/json-schemas/src/client-events-api-requests.json
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "Ably API Requests Client Event",
   "type": "object",
-  "description": "A message sent on the clientEvents:apiRequests metachannel describing a sampled client api request.",
+  "description": "A message sent on the clientEvents:apiRequests metachannel describing a sampled client API request.",
   "properties": {
     "host": {
       "type": "string",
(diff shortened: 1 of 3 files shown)


Jmgr:
See comment https://ably.atlassian.net/browse/REA-980?focusedCommentId=24385

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ably API Requests Client Event",
+  "type": "object",
```
lmars, line 4: Let's also add a description, for example:

```
"description": "A message sent on the clientEvents:apiRequests metachannel describing a sampled client api request."
```

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,41 @@
...
+  "properties": {
+    "host": {
+      "type": "string",
+      "description": "the host the request was made to"
```
lmars, line 8: Let's format these descriptions as proper sentences (i.e. start with a capital letter, and end with a full stop).

Also, I assume this is the HTTP Host header from the request, in which case I'd alter the description to:

```
The HTTP Host header of the API request.
```

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "requestId": {
+      "type": "string",
+      "description": "the unique ID of the request that Ably can use to correlate a request event with internal logs, if necessary"
+    },
+    "region": {
+      "type": "string",
+      "description": "the region the request was made in"
```
lmars, line 16:
```
The region the request was processed in.
```

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "headers": {
+      "type": "object",
+      "description": "the headers sent with the request"
```
lmars, line 20:
```
The HTTP headers of the API request.
```

json-schemas/src/client-events-api-requests.json
lmars: And I just looked at a sample message, and it seems this is in fact a site, not a region (i.e. it has a `-A` suffix):

```
"region": "eu-central-1-A"
```

Should we change it to `site` now before we publish this schema?

json-schemas/src/client-events-api-requests.json
lmars: Also, it's worth noting that the HTTP header keys are normalised to all lower case (i.e. `user-agent` rather than `User-Agent`)

json-schemas/src/client-events-api-requests.json
Jmgr: > Let's format these descriptions as proper sentences (i.e. start with a capital letter, and end with a full stop).

For some reason I was convinced that the convention for JSON Schema was to drop both, but I cannot find any evidence of that now. 

> Also, I assume this is the HTTP Host header from the request, in which case I'd alter the description to:

Could be, I'll check.

json-schemas/src/client-events-api-requests.json
Jmgr: That would make sense, I'll also change the code in realtime to do that.

json-schemas/src/client-events-api-requests.json
Jmgr: Yes. Do you mean that should be included in the description? Or that the code should enforce that by making all headers lower case?

json-schemas/src/client-events-api-requests.json
lmars: My reasoning is that I expect these descriptions to appear in both documentation (e.g. [here](https://ably.com/documentation/realtime/metachannels#connections) and in UIs when users interact with these messages, where I would expect them to be proper sentences.

json-schemas/src/client-events-api-requests.json
lmars: I think it should be in the description to inform the user, but no enforcement necessary since HTTP header names are case insensitive.

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,42 @@
...
+      "type": "string",
+      "description": "The unique ID of the request that Ably can use to correlate a request event with internal logs, if necessary."
+    },
+    "site": {
+      "type": "string",
+      "description": "The site the request was processed in."
+    },
+    "headers": {
+      "type": "object",
+      "description": "The HTTP headers sent with the request. Keys should be lower case."
```
lmars, line 21: "should" is too strong, it's not a requirement, it's just the convention we use:

```
Header names are lower case (e.g. 'accept', 'content-type', 'user-agent', etc.)
```

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,42 @@
...
+    },
+    "query": {
+      "type": "object",
+      "description": "The details of the parsed request query, including information such as the request format."
+    },
+    "path": {
+      "type": "string",
+      "description": "The path of the endpoint called in the request, for example /channels/{channel-name}/messages."
```
lmars, line 29: "query" and "path" should probably have similar descriptions, they currently feel a bit wordy and inconsistent. How about:

- query: "The parsed query component of the API request URL (e.g. {\"format\":\"msgpack\"})."
- path: "The path component of the API request URL (e.g. /channels/{channel-name}/messages)."

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,42 @@
...
+    },
+    "channels": {
+      "type": "array",
+      "description": "A list of channels the request was made against."
```
lmars, line 33: I don't like the term "made against", how about "references"?

"The list of channels the API request references."

json-schemas/src/client-events-api-requests.json
```diff
@@ -0,0 +1,42 @@
...
+    },
+    "error": {
+      "type": "object",
+      "description": "The details of any error encountered whilst making the request. It includes an error message, error code and HTTP status code."
```
lmars, line 37: Given this error object has a well defined structure (as indicated here in the description), should it have a schema? Maybe it should be a separate schema that's referenced here?

json-schemas/src/client-events-connections.json
```diff
@@ -0,0 +1,50 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ably Connections Client Event",
+  "type": "object",
+  "description": "A message sent on the clientEvents:connections metachannel describing a sampled client connection.",
+  "properties": {
+    "host": {
+      "type": "string",
+      "description": "The host the connection was made to."
```
lmars, line 9: This should be consistent with the descriptions in the api requests schema.

json-schemas/src/client-events-api-requests.json
Jmgr: > Given this error object has a well defined structure (as indicated here in the description), should it have a schema?

Good point.

> Maybe it should be a separate schema that's referenced here?

Using an external schema will be painful to implement in unit tests, so I'd rather be for adding the three missing properties there.

json-schemas/src/client-events-connections.json
```diff
@@ -0,0 +1,65 @@
...
+      "type": "string",
+      "description": "The HTTP Host header sent with the connection."
+    },
+    "requestId": {
+      "type": "string",
+      "description": "The unique ID of the request that Ably can use to correlate a connection event with internal logs, if necessary."
+    },
+    "site": {
+      "type": "string",
+      "description": "The site the connection was made to."
```
lmars, line 17: This should be the same as the description in the api requests schema.

json-schemas/src/client-events-connections.json
```diff
@@ -0,0 +1,65 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ably Connections Client Event",
+  "type": "object",
+  "description": "A message sent on the clientEvents:connections metachannel describing a sampled client connection.",
+  "properties": {
+    "host": {
+      "type": "string",
+      "description": "The HTTP Host header sent with the connection."
```
lmars, line 9: You don't really _send_ a connection. How about:

"The HTTP Host header used when establishing the connection."

(and similarly in other descriptions, replace "making" or "sending" with "establishing" a connection)

Jmgr: I thought it was implied that this meant "sent with the connection request", but I'm ok with your suggestion.

lmars (approved):

PR #124, opened 2021-09-30
Add versions for clientEvent JSON schemas

lmars:
As per the [JSON schema docs](https://github.com/ably/ably-common/tree/main/json-schemas#versioning), each schema needs an entry in `versions.json` for it to be published, although we forgot to do that in https://github.com/ably/ably-common/pull/118, so this PR adds those versions.

I've also updated the JSON schema docs to include instructions for AWS access.

### DIFF
--- a/json-schemas/README.md
+++ b/json-schemas/README.md
@@ -20,6 +20,17 @@ publish time (see below).
 
 ## Publish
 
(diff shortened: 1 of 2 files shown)


Jmgr (approved):
LGTM

PR #131, opened 2021-12-09
Remove the duplicates slugs from FAQ error urls

denissellu:
Old FAQ urls for error pages has duplicates pages in hubspot and "-1" was added to the slugs, since the hubspot pages were clean up the urls needed to be updated


lmars (approved):

markcoxsmith (approved):
Thanks Denis. All looks good

PR #132, opened 2022-01-04
Update agents.json

AndyNicks:
Added Flutter & kafka-connect to "ablyLibMappings" list

### DIFF
--- a/protocol/agents.json
+++ b/protocol/agents.json
@@ -232,6 +232,8 @@
     "php-laravel": "ably-php/${version} laravel",
     "python": "ably-python/${version}",
     "ruby": "ably-ruby/${version}",
-    "ruby-rest": "ably-ruby/${version} ably-ruby-rest"
+    "ruby-rest": "ably-ruby/${version} ably-ruby-rest",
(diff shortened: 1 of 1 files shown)


protocol/agents.json
```diff
@@ -232,6 +232,8 @@
     "php-laravel": "ably-php/${version} laravel",
     "python": "ably-python/${version}",
     "ruby": "ably-ruby/${version}",
-    "ruby-rest": "ably-ruby/${version} ably-ruby-rest"
+    "ruby-rest": "ably-ruby/${version} ably-ruby-rest",
+    "flutter": "ably-flutter/${version}",
+    "kafka-connect": "kafka-connect-ably/${version}"
```
lmars, line 237: As described [this section on ablyLibMappings](https://github.com/ably/ably-common/tree/main/protocol#ablylibmappings), only libraries which continue to send the deprecated `X-Ably-Lib` header should be added to this list, but I don't believe either flutter or kafka connect do so?

AndyNicks: My bad. You are completely right

PR #136, opened 2022-03-21
Conform Tool Versions between CI and Local Development

QuintinWillison:
This had been bugging me for a while, as:

- Our cloud builds were at the whim of GitHub runtime environment upgrades:
  - in respect of Node.js (/ `npm`) version
  - in respect of Golang patch version
- When developing locally I was never sure which version of tools I _should_ be developing with... personally I like to use [ASDF](https://github.com/asdf-vm/asdf) to manage and conform this, hence update to `.tool-versions`

This pull request ensures that we have a single, canonical location for tool versions for both Node.js and Golang required by CI/CD within this repository.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,7 +12,7 @@ jobs:
 
       - id: tool-versions
         run: |
-          run: echo "::set-output name=nodejs::$(grep nodejs .tool-versions | cut -d ' ' -f 2)"
+          run: echo "::set-output name=nodejs::$(sed -nr 's/nodejs ([0-9]+)/\1/p' .tool-versions)"
 
       - uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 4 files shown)


lmars (approved):

owenpearson (approved):
LGTM. I'd _maybe_ consider using https://github.com/marocchino/tool-versions-action since that `sed` command looks a bit opaque.

Peter-Maguire (approved):

QuintinWillison:
Thanks, @owenpearson ... I did take a quick look at the third-party action you refer to, however every third party action we add to any repository comes with additional risks that fall outside of our control, and this is _just_ a 'one liner', so on balance I preferred something internal and controlled.

Perhaps we could consider our own action in due course... maybe even as part of a conformed standard set of workflow pre-ambles that we use across our repositories. 🤔

PR #140, opened 2022-03-24
Lint markdown files

QuintinWillison:
We've been using the `markdownlint-cli2` tool [over in `ably/engineering`](https://github.com/ably/engineering/blob/main/package.json#L6) for a few months now without any problems, so I'm bringing that here now.


owenpearson (approved):
🚀

Peter-Maguire (approved):

PR #144, opened 2022-04-11
Add appId to app-stats JSON schema

lmars:
We plan to use the app-stats schema for internal analytics purposes where it will be useful for app-stats objects to have an appId field.

Since this includes the addition of an optional field, I've made a patch version bump to 0.0.2.


ruimffl (approved):
looks good to me

QuintinWillison (approved):

PR #146, opened 2022-05-16
Update client-events JSON schemas

lmars:
Specifically, the `headers` and `query` definitions in both `client-events-api-requests.json` and `client-events-connections.json`.

I've chosen to only document the HTTP headers which are specific to Ably, the rest are best described by searching the Internet 🙂

### DIFF
--- a/json-schemas/src/client-events-api-requests.json
+++ b/json-schemas/src/client-events-api-requests.json
@@ -18,11 +18,35 @@
     },
     "headers": {
       "type": "object",
-      "description": "The HTTP headers of the API request. Header names are lower case (e.g. 'accept', 'content-type', 'user-agent', etc.)."
+      "description": "The HTTP headers of the API request. Header names are lower case (e.g. 'accept', 'content-type', 'user-agent', etc.).",
+      "properties": {
+        "ably-agent": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


ruimffl (approved):
Looks good. Will propagate the changes in the dbt docs. Thanks for clarifying the definitions

SimonWoolf:
Not sure about this. You've picked four querystring params which are common ones and put them in the schema. But `query` is just whatever querystring parameters the sdk sends, could in general be anything, doesn't picking four imply that those are all there is? Isn't there a way in json-schema to mark a field as having type `Object<string, string>` or something?

lmars:
> including any field other than those would not be valid?

JSON schema is permissive by default, it's only if `additionalProperties` is set to `false` that validation will fail for keys which aren't explicitly listed, see [here](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties).

So, the intention here was to validate that, if any of the four keys explicitly listed are present, that they are strings and have the described semantics, I'm not sure how else we'd achieve that.

> Isn't there a way in json-schema to mark a field as having type Object<string, string> or something?

We could set `"additionalProperties": { "type": "string" }` to validate that, should any additional keys be present, that their values are strings.

PR #155, opened 2022-06-22
Add app setup with revocable tokens enabled

ikbalkaya:
This adds an additional file so that simulations app can be setup with `revocableTokens = true`

Required for tests here: https://github.com/ably/ably-cocoa/pull/1430


test-resources/test-app-setup-for-revocation.json
```diff
@@ -0,0 +1,75 @@
...
+  "post_apps": {
+    "limits": {
+      "presence": {
+        "maxMembers": 250
+      }
+    },
+    "keys": [
+      {
+        "revocableTokens": true
+      },
```
lmars, line 12: Can we not just add this as a new key in `test-app-setup.json`?

ikbalkaya: @lmars Yes of course, It's the first time using this file, so I must have thought there was a single app key - I added a separate key 591e9cd199f6c213fd4f9d4a9422283fb789c3a6 and removed file 3949a8d4b93a0c4ba5be07dcf160e3d3417cc9e4

lmars (approved):

PR #163, opened 2022-08-08
Add 40006 error code redirect

denissellu:
This will add a redirect for the 40006 error code to https://faqs.ably.com/error-code-40006-invalid-connection-id


owenpearson (approved):
LGTM

PR #165, opened 2022-08-25
Add client events connection's token credentials

Jmgr:
This is the ably-common changes for [#4269](https://github.com/ably/realtime/pull/4269).

This adds client events connection's token credentials information to the schema.

### DIFF
--- a/json-schemas/src/client-events-connections.json
+++ b/json-schemas/src/client-events-connections.json
@@ -76,7 +76,7 @@
       "type": "integer",
       "description": "The duration the connection was open for in milliseconds."
     },
-    "credentials": {
+    "credential": {
       "type": "object",
       "description": "The connection's token credentials.",
       "properties": {
@@ -93,7 +93,7 @@
           "description": "The credential type: either 'key', 'ablyToken', 'ablyJWT', or 'ablyJWE'."
         }
       },
-      "required": ["hasUserClaims", "type"]
+      "required": ["hasUserClaims", "keyName", "type"]
     },
     "error": {
       "type": "object",
--- a/json-schemas/src/client-events-connections.json
+++ b/json-schemas/src/client-events-connections.json
@@ -76,6 +76,25 @@
       "type": "integer",
       "description": "The duration the connection was open for in milliseconds."
     },
+    "credentials": {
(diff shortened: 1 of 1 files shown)


json-schemas/src/client-events-connections.json
```diff
@@ -76,6 +76,25 @@
       "type": "integer",
       "description": "The duration the connection was open for in milliseconds."
     },
+    "credentials": {
```
lmars, line 79: There is only one credential in a request, so should this just be `credential`?

json-schemas/src/client-events-connections.json
```diff
@@ -76,6 +76,25 @@
...
+      "type": "object",
+      "description": "The connection's token credentials.",
```
lmars, line 81:
```suggestion
      "description": "Details of the credential used to authenticate the client.",
```

I say "Details of" since it isn't the actual credential, just details about it.

json-schemas/src/client-events-connections.json
```diff
@@ -76,6 +76,25 @@
...
+      "properties": {
+        "hasUserClaims": {
+          "type": "boolean",
+          "description": "Does this token contain user claims."
+        },
+        "keyName": {
+          "type": "string",
+          "description": "The token's key name."
```
lmars, line 89:
```suggestion
          "description": "The credential's key name."
```

json-schemas/src/client-events-connections.json
```diff
@@ -76,6 +76,25 @@
...
+        },
+        "type": {
+          "type": "string",
+          "description": "The credential type: either 'key', 'ablyToken', 'ablyJWT', or 'ablyJWE'."
+        }
+      },
+      "required": ["hasUserClaims", "type"]
```
lmars, line 96: Why isn't `keyName` required?

Jmgr: According to the TS definition `this.authParams.getKeyUsername()` can return a string or null, however looking at the implementation it seems that this is not correct: a string is returned in all cases. I'll fix that.

cruickshankpg (approved):
LGTM

PR #167, opened 2022-09-06
Fix name of channel lifecycle events schema

mschristensen:
In order to match the naming conventions used by the stored procedures
developed by the data team, we need to remove the -events suffix from
this schema definition.


mschristensen:
@lmars updating the schema according to the changes requested [here](https://ably.atlassian.net/browse/BIA-267)

lmars (approved):

PR #171, opened 2022-10-31
json-schemas: add account stats message schema

mschristensen:
This PR adds a new schema for account-level statistics as part of [REA-1414](https://ably.atlassian.net/browse/REA-1414).

The statistics we will send are the same as the app stats, the aggregation is just being performed at account level. However I have added this as a separate schema to keep aligned with the existing approach and in expectation of divergence of account and app stats in the future.

### DIFF
--- a/json-schemas/src/account-stats.json
+++ b/json-schemas/src/account-stats.json
@@ -1,0 +1,852 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ably Account Statistics Message",
+  "type": "object",
+  "properties": {
+    "accountId": {
+      "type": "string",
+      "description": "The ID of the Ably account the statistics are for."
+    },
+    "intervalId": {
+      "type": "string",
+      "description": "time period for statistics in date-time format yyyy-mm-dd:hh:mm:ss"
+    },
+    "unit": {
+      "type": "string",
+      "description": "Unit of time for these statistics from the intervalId forwards.",
+      "pattern": "^minute$"
+    },
+    "schema": {
+      "type": "string",
+      "description": "URI to this schema."
+    },
+    "entries": {
+      "type": "object",
+      "description": "Statistics entries.",
+      "properties": {
+        "messages.all.all.count": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total message count."
+        },
+        "messages.all.all.data": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total message size."
+        },
+        "messages.all.all.uncompressedData": {
+          "type": "number",
+          "inclusiveMinimum": 0,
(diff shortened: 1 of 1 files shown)


mschristensen:
Corresponding realtime draft PR is here: https://github.com/ably/realtime/pull/4439

json-schemas/src/account-stats.json
```diff
@@ -0,0 +1,802 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ably Account Statistics Message",
+  "type": "object",
+  "properties": {
+    "accountId": {
```
ruimffl, line 6: this is the account API id right?

json-schemas/src/account-stats.json
```diff
@@ -0,0 +1,802 @@
...
+    "unit": {
+      "type": "string",
+      "description": "Unit of time for these statistics from the intervalId forwards.",
+      "pattern": "^minute$"
+    },
+    "schema": {
+      "type": "string",
+      "description": "URI to this schema."
+    },
+    "entries": {
```
ruimffl, line 23: can I assume 1 single entry per account stats entry, i.e. no need to flatten an array, similarly to the App stats?

mschristensen: correct

json-schemas/src/account-stats.json
```diff
@@ -0,0 +1,802 @@
...
+        "push.notifications.all": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of Push notifications."
+        },
+        "push.directPublishes": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of direct publishes."
+        }
```
SimonWoolf, line 796: Account stats has a few fields that aren't in app stats but which we probably want in here. The xchg ones I think we can ignore for now (unless/until we decide what's happening with api streamer), but we do want `peakRates` here.

so that's the following new entries:
peakRates.messages
peakRates.apiRequests
peakRates.tokenRequests
peakRates.connections
peakRates.channels
peakRates.reactor.httpEvent
peakRates.reactor.amqp
peakRates.reactor.externalQueue
peakRates.reactor.webhook
peakRates.pushRequests

ruimffl (approved):
Thanks for the clarifications @mschristensen and @SimonWoolf.

When you go to the part of pushing the files to S3, it can be sent to 
`s3://ably-analytics-data-lake-sandbox/ably-api-events/account_stats/`
and 
`s3://ably-analytics-data-lake-production/ably-api-events/account_stats/`

json-schemas/src/account-stats.json
mschristensen: Thanks Simon - I've added these now

SimonWoolf (approved):

PR #177, opened 2022-12-29
Dotnet agent headers

sacOO7:
- Removed uwp from ably-lib mappings since its deprecated and removed as a part of https://github.com/ably/ably-dotnet/pull/1101
- Refactored xamarin variants since it's tightly coupled with android/iOS, also avoid duplication under agents mappings.
- Removed versioning for dotnet variants since it's impossible to find stable API that can give versioning across different dotnet variants.


owenpearson (approved):

owenpearson (changes requested):
I'm not sure if I agree with the change from `xamarin android` to `xamarin-android` because could be useful to see how many clients are running on android/ios across our SDKs. what's the reason for the change?

Also, I notice we're removing unity as an agent here but I don't see any reasons why, could you add something to the description to make it clear?

sacOO7:
We don't support unity as an agent header yet. Based on my experience working with unity, it would be impossible to add it at this point. So, when we add it to `unity`, then only we should add it here.

sacOO7:
@owenpearson I don't have a problem keeping them separate. I kept it together because of two reasons.
1. Current iOS/android agents under `agents.json` are versioned. I tried to find a way to get active android/iOS versions based on xamarin. But. there is no stable API available for it.
2. Xamarin is a framework made especially for android/iOS, there is no independent identity for it.

owenpearson (approved):

protocol/agents.json
```diff
@@ -208,11 +203,10 @@
     "cocoa.macos": "ably-cocoa/${version} macOS/0.0.0",
     "cocoa.tvos": "ably-cocoa/${version} tvOS/0.0.0",
     "dotnet": "ably-dotnet/${version}",
-    "dotnet.framework": "ably-dotnet/${version} dotnet-framework/0.0.0",
-    "dotnet.uwp": "ably-dotnet/${version} uwp",
```
lmars, line 212: Are you sure this was never used in any previously released SDK?

sacOO7: We have removed support for this since ably-dotnet 1.2.7 (https://github.com/ably/ably-dotnet/pull/1101).  As far as agent headers are concerned, there's no point in having the mapping since we are not emitting `uwp` as a part of agent-headers ( it doesn't exist in the first place).

lmars: > there's no point in having the mapping since we are not emitting uwp as a part of agent-headers ( it doesn't exist in the first place).

Ok thanks, I just wanted to clarify that 👍

lmars (approved):

PR #186, opened 2023-03-16
protocol: Add Control API Ably-Agent entries

lmars:


lmars:
> The validation is based on a regex to grab numeric values. Do you think we need to change that to cover `version=dev`

Ah I was unaware of that, so if it's simple enough to allow `dev` then let's do that, but if not we could use a number to indicate dev in the terraform provider instead (e.g. `0.0.0` or `-1`)

PR #188, opened 2023-03-20
Add faq for error code 10000

jamienewcomb:
Another code we have been firing but we were not directing customers to an FAQ. this fixes that

Please merge once approved


jamienewcomb:
Will start this one again

PR #189, opened 2023-03-20
Add faq for error 10000

jamienewcomb:
This error was firing over the weekend and we currently don't have an faq linked. This fixes that 

Please merge once approved


lmars (approved):
Attention! There was no error 😅

PR #194, opened 2023-04-04
agents: add ably db connector

andydunstall:
Adds the ably database connector agent


mschristensen (approved):
Approving, but also needs one from Rui :)

ruimffl (approved):
looks good to me. I'll update the parser on our end. Thanks for implementing this :)

PR #200, opened 2023-05-15
WIP: sdk issue form template

AndyTWF:


templates/sdk-github-issue-template.yaml
```diff
@@ -0,0 +1,48 @@
...
+body:
+  - type: input
+    id: library-version
+    attributes:
+      label: Library version
+      description: What version of the library are you using?
+      placeholder: ex. v1.2.21
+    validations:
+      required: true
+  # Can remove this field for most SDKs, but useful for JS
```
owenpearson, line 14: Honestly I think this wouldn't even be useful for JS (and certainly not for js v2 😄). I would suggest replacing this with a "Which platform are you experiencing this issue on" (runtime version, browser/react-native/nodejs, etc) which would probably be useful for every SDK.

templates/sdk-github-issue-template.yaml
```diff
@@ -0,0 +1,48 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug", "triage"]
```
owenpearson, line 4: I'm not against using the "triage" label but if we're gonna go for that we ought to add it to https://github.com/ably/engineering/blob/main/sdk/github.md#labels and document some process about how we use it - i guess that would just be "remove the label once the bug is prioritised"

templates/sdk-github-issue-template.yaml
```diff
@@ -0,0 +1,48 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behaviour
+    attributes:
+      label: What did you expect to happen?
+      description: If things were working as you'd expect, what would have you seen instead?
+    validations:
+      required: true
+  - type: textarea
```
owenpearson, line 37: My personal preference would be to  merge these two fields into an "additional information" field which suggests sharing a minimal repro or debug logs

owenpearson (changes requested):

PR #203, opened 2023-06-07
attachments: add attachments events schema

mschristensen:
This adds a new event schema definition for attachment events to support [usage analytics](https://ably.atlassian.net/wiki/spaces/ENG/pages/2614198277/ADR94+User+analytics+pipeline) use cases.

See the corresponding realtime PR for more info.


json-schemas/src/attachments.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "channelName": {
+      "type": "string",
+      "description": "The name of the channel"
+    },
+    "qualifier": {
+      "type": "string",
+      "description": "The qualifier used in the channel spec"
+    },
+    "qualifierParam": {
```
lmars, line 18: Would it be useful to also include the qualifier query (i.e. the `rewind=1` in `[?rewind=1]foo`)?

json-schemas/src/attachments.json
```diff
@@ -0,0 +1,25 @@
...
+      "type": "string",
+      "description": "The parameter for the qualifier used in the channel spec"
+    }
```
lmars, line 21: Perhaps include the requested [channel params](https://ably.com/docs/realtime/channels/channel-parameters/overview) and [mode](https://faqs.ably.com/what-are-channel-mode-flags-and-how-do-i-use-them)?

mschristensen:
> Are we able to query this data by app and/or account id? I'm curious if these events are decorated with that data during ingestion to Snowflake?

@emptyhammond yes, the app ID is included on the event but it is not included in this schema (which is the same as the other analytics events, which also include it but don't show it in the schema; there is a misalignment between customer-facing meta-channel schemas and those used internally, see: https://ably.atlassian.net/browse/REA-1370).

The account id gets pulled in during the transformation step, see the discussion on the realtime PR.

json-schemas/src/attachments.json
mschristensen: Haven't included mode, how useful is that really?

json-schemas/src/attachments.json
lmars: @mschristensen did you intend to add this?

paddybyers:
Is this ready to merge?

lmars (approved):

paddybyers:
Does this actually need to be landed in order to get the events shipped? I assume not

lmars:
> Does this actually need to be landed in order to get the events shipped?

I don't believe so, no.

mschristensen:
Following this conversation I'll merge this as-is, we can add rewind and others later if needed.

PR #208, opened 2023-07-14
analytics: add connection and attachment message rates

andydunstall:
Adds message statistics to connection and attachment message rates as described in https://ably.atlassian.net/browse/REA-317:

```
Collect the following stats on a per-connection basis, and emit with the connection event via the analytics service:
* total message count (break down by inbound/outbound and messages/presence)
* total message data (break down by inbound/outbound and messages/presence)
* message rate high water mark(s)

Collect the following stats on a per-attachment basis, and emit with the attachment event via the analytics service:
* total message count (break down by inbound/outbound and messages/presence)
* total message data (break down by inbound/outbound and messages/presence)
* (for filtered subscription attachments) filter acceptance rate (or total processed + total matching the filter)
```


json-schemas/src/client-events-connections.json
```diff
@@ -77,6 +77,88 @@
       "type": "integer",
       "description": "The duration the connection was open for in milliseconds."
     },
+    "messages": {
```
lmars, line 80: If we have these numbers on a per-attachment basis, then we can get the per-connection numbers by summing them, so is it redundant to also include these in connection events?

andydunstall: Its easy to include both from the realtime side - so I guess its up to @ruimffl which is easier from the data side?

ruimffl: If I understand correctly duration is already included in the connection events and we are using it to compute average and min/max duration when aggregating by day. It would be easier to keep it there instead of having to join to the attachment events. Hope I've answer the question (?)

andydunstall:
@ruimffl Is this schema ok? The realtime side is in review so it would be good to get approval on this - thanks

ruimffl (approved):
It all looks good to me @andydunstall 

The data will only be available in Snowflake after we migrate our ingestion process from Json to Parquet that should happen next week but there is no issues with this being deployed before.

PR #215, opened 2023-08-25
[MMB-44] Assign error code range for spaces

surminus:
I've put in a placeholder for the actual specification - not sure where we want to add this?

### DIFF
--- a/protocol/README.md
+++ b/protocol/README.md
@@ -39,17 +39,26 @@ Codes in [errors.json](errors.json) fall into these ranges:
 | 80000 | connection-related |
 | 90000 | channel-related |
 
-Codes from `100000` are defined outside of scope of this repository.
-In other words, these higher ranges are reserved for use by specific Ably projects and ecosystems, where these codes are unknown to the core Ably service:
+Codes above `100000` are reserved for use by specific Ably projects and ecosystems, where these codes are unknown to the core Ably service. Definitions for these ranges are defined below.
 
-| From | To | Project / Ecosystem | Location of Definitions |
-| ---- | -- | ------------------- | ----------------------- |
-| 100000 | 100999 | Asset Tracking | [Specification: Error codes](https://github.com/ably/ably-asset-tracking-common/tree/main/specification#error-codes) |
+#### Asset Tracking
 
-The table above is our canonical location for:
+Codes for Asset Tracking are maintained externally.
 
-- specifying these ranges
-- delegating the definitions of codes that sit within these ranges to elsewhere
+| From | To | Location of Definitions |
+| ---- | -- | ----------------------- |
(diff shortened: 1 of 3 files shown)


protocol/README.md
```diff
@@ -45,6 +45,7 @@ In other words, these higher ranges are reserved for use by specific Ably projec
 | From | To | Project / Ecosystem | Location of Definitions |
 | ---- | -- | ------------------- | ----------------------- |
 | 100000 | 100999 | Asset Tracking | [Specification: Error codes](https://github.com/ably/ably-asset-tracking-common/tree/main/specification#error-codes) |
+| 101000 | 101999 | Spaces | [Specification: Error codes](https://github.com/ably-labs/spaces/tree/main/specification#error-codes) |
```
lmars, line 48: Thinking about it, I think we should just put the error codes in this repo, maybe just in `errors.json` with the others? I don't see a reason to maintain them separately tbh.

surminus: We can do this, but the docs above this table no longer makes sense when it states that:

> Codes from `100000` are defined outside of scope of this repository.
> In other words, these higher ranges are reserved for use by specific Ably projects and ecosystems, where these codes are unknown to the core Ably service

I can change the wording slightly so other SDKs can re-use this repository.

surminus: I've updated in 5fb8a0434a4a010e599273863a3ae4c102868f68.

I've also included a link to FAQs in `errorsHelp.json`... but I'm guessing this doesn't exist yet. Is this something we need to add?

protocol/README.md
```diff
@@ -39,17 +39,26 @@ Codes in [errors.json](errors.json) fall into these ranges:
 | 80000 | connection-related |
 | 90000 | channel-related |
 
-Codes from `100000` are defined outside of scope of this repository.
-In other words, these higher ranges are reserved for use by specific Ably projects and ecosystems, where these codes are unknown to the core Ably service:
+Codes from `100000` are are reserved for use by specific Ably projects and ecosystems, where these codes are unknown to the core Ably service. Definitions for these ranges are defined below.
```
lmars, line 42:
```suggestion
Codes above `100000` are reserved for use by specific Ably projects and ecosystems, where these codes are unknown to the core Ably service. Definitions for these ranges are defined below.
```

lmars (approved):
Minor comment, otherwise LGTM

protocol/README.md
lmars: > I'm guessing this doesn't exist yet. Is this something we need to add?

They need to be added but I'm not sure by who, see [this Slack discussion](https://ably-real-time.slack.com/archives/C04B5EKDUKC/p1692878438128949).

PR #216, opened 2023-08-30
add react-hooks-time-ping agent

owenpearson:
This agent is used by react-hooks to differentiate agents coming from the libraries `/time` request upon init from agents coming from actual usage of the library when the agent is succesfully added before a connection is made.

### DIFF
--- a/protocol/agents.json
+++ b/protocol/agents.json
@@ -316,6 +316,11 @@
       "type": "wrapper"
     },
     {
+      "identifier": "react-hooks-time-ping",
+      "versioned": true,
+      "type": "wrapper"
+    },
+    {
       "identifier": "ably-spaces",
       "versioned": true,
       "type": "wrapper"


AbdullahWins (commented):
Good Job!

sacOO7:
> This agent is used by react-hooks to differentiate agents coming from the libraries `/time` request upon init from agents coming from actual usage of the library when the agent is succesfully added before a connection is made.

How is it different from `react-hooks` agent header? Like for `time` api, we will also get `react-hooks` header right

owenpearson:
@sacOO7 due to the recommended way to use our react-hooks package, typically the connection is established before we have a chance to add the `react-hooks` agent so instead we send an arbitrary request to `/time` upon init - this agent is the agent that is used for that arbitrary request to differentiate it from those made from a genuine usage of the library.

owenpearson:
@sacOO7 yes but the most common usage of react-hooks is just to use realtime channel features, rather than the REST API. Since the connection is established before we can append the `react-hooks` agent there would be no way to distinguish between a normal ably-js connection and a connection created with react-hooks. To solve this, we send a special `/time` request with this new agent to make sure that we're aware when someone is using the `react-hooks` library.

sacOO7:
> @sacOO7 yes but the most common usage of react-hooks is just to use realtime channel features, rather than the REST API. Since the connection is established before we can append the `react-hooks` agent there would be no way to distinguish between a normal ably-js connection and a connection created with react-hooks. To solve this, we send a special `/time` request with this new agent to make sure that we're aware when someone is using the `react-hooks` library.

So, this is a special `time` request, rather than using `ablyrest.time()` api. No issues, I was just trying to understand, if additional agent header was really required.

stmoreau (approved):

PR #218, opened 2023-09-15
Json-schema for epoch events: fix namsepace identified field name

SimonWoolf:


ruimffl:
Thanks for adding me @SimonWoolf, I'll replicate the change in the analytics view. 
Should the new field `identified` map to the new key `identified` only, or should it map to `authenticated`  when `identified` is not present, i.e. before this change is live?

SimonWoolf:
> Should the new field identified map to the new key identified only, or should it map to authenticated when identified is not present, i.e. before this change is live?

it's the same key, just renamed. the corresponding change was deployed in realtime last week. (doesn't hugely matter if the data on this isn't available before last week, relatively unlikely anyone's going to notice/care)

some more context: 'authenticated' was renamed 'identified' [_back in 2015_](https://github.com/ably/docs/issues/49)...

PR #219, opened 2023-10-05
protocol: add missing code to errors.json

Zariel:
Add error code 50305.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -103,6 +103,7 @@
   "50320": "Active Traffic Management: traffic for this cluster is being temporarily redirected to a backup service",
 
   "50330": "request reached the wrong cluster; retry (used during dns changes for cluster migrations)",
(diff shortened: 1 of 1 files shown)


protocol/errors.json
```diff
@@ -103,6 +103,7 @@
   "50320": "Active Traffic Management: traffic for this cluster is being temporarily redirected to a backup service",
 
   "50330": "request reached the wrong cluster; retry (used during dns changes for cluster migrations)",
+  "50305": "Ably's routing layer was unable to service this request",
```
lmars, line 106: Should we add [50405](https://github.com/ably/realtime/blob/main/go/realtime/roles/frontdoor/router/request_handler.go#L259) and [50205](https://github.com/ably/realtime/blob/main/go/realtime/roles/frontdoor/router/request_handler.go#L262) too?

PR #226, opened 2023-11-30
Add conflict error code

JoaoDiasAbly:


protocol/errors.json
```diff
@@ -79,6 +79,8 @@
 
   "40500": "method not allowed",
 
+  "400900": "conflict",
```
AndyTWF, line 82: Looks like there's an extra zero here?

JoaoDiasAbly: Lack of attention on my part, thanks

AndyTWF (approved):

PR #227, opened 2023-12-12
Remove 'spaces-*-client' agent entries

lmars:
They are no longer used.


ruimffl (approved):
👍 thanks for the heads-up

PR #228, opened 2023-12-13
Error code for token auth only

AndyTWF:
Adds an error code to indicate that the operation in question may only be performed by a client that is using token auth.


AndyTWF:
Actually @lmars - would this code be better suited in the 40000 range, similar to 40012 which is "invalid client id"?

lmars (approved):

lmars:
@AndyTWF I think 401xx is fine, I don't think it's a 400xx (i.e. there is nothing invalid in the request), if anything it could be a 403xx for Forbidden.

paddybyers:
> if anything it could be a 403xx for Forbidden.

401xx pls

PR #245, opened 2024-04-25
JSON schemas: fix schema for push stats, which was completely wrong

SimonWoolf:
REA-1898


ruimffl (approved):
LGTM, thank you for fixing it so quickly

owenpearson (approved):

PR #255, opened 2024-07-25
App account stats schemas: add lots of missing entries, fix 'processed' entries, update terminology

SimonWoolf:
REA-1986

Lots of failed and refused entries were missing, probably because they were missing from https://ably.com/docs/metadata-stats/stats and I think this was probably originally created from that.

The `processed` stats (delta generation) were also completely wrong.

Also updated the terminology to our current stuff (we don't say "reactor" any more, etc)

### DIFF
--- a/json-schemas/src/account-stats.json
+++ b/json-schemas/src/account-stats.json
@@ -42,7 +42,7 @@
         "messages.all.all.failed": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total number of messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as rejected by an external integration target, or a service issue on Ably's side)"
+          "description": "Total number of messages that failed (which did not succeed for some reason other than Ably explicitly refusing it, such as rejected by an external integration target, or a service issue on Ably's side)"
         },
         "messages.all.all.refused": {
           "type": "number",
@@ -62,12 +62,12 @@
         "messages.all.messages.uncompressedData": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total number of messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as rejected by an external integration target, or a service issue on Ably's side)"
+          "description": "Total number of messages that failed (which did not succeed for some reason other than Ably explicitly refusing it, such as rejected by an external integration target, or a service issue on Ably's side)"
         },
         "messages.all.messages.failed": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total number of messages excluding presence messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as rejected by an external integration target, or a service issue on Ably's side)"
+          "description": "Total number of messages excluding presence messages that failed (which did not succeed for some reason other than Ably explicitly refusing it, such as rejected by an external integration target, or a service issue on Ably's side)"
         },
         "messages.all.messages.refused": {
           "type": "number",
@@ -92,7 +92,7 @@
         "messages.all.messages.failed": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total number of presence messages excluding presence messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as rejected by an external integration target, or a service issue on Ably's side)"
+          "description": "Total number of presence messages excluding presence messages that failed (which did not succeed for some reason other than Ably explicitly refusing it, such as rejected by an external integration target, or a service issue on Ably's side)"
         },
         "messages.all.messages.refused": {
           "type": "number",
@@ -117,7 +117,7 @@
         "messages.inbound.realtime.all.failed": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total number of inbound realtime messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as a service issue on Ably's side)"
+          "description": "Total number of inbound realtime messages that failed (which did not succeed for some reason other than Ably explicitly refusing it, such as a service issue on Ably's side)"
         },
         "messages.inbound.realtime.all.refused": {
           "type": "number",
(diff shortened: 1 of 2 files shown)


paddybyers (commented):
A couple of editorial suggestions.

owenpearson (approved):
looks good, minor comment:

json-schemas/src/account-stats.json
```diff
@@ -27,17 +27,27 @@
...
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total message size."
+          "description": "Total message size of all messages successfully sent, summed over all message types and transports."
         },
         "messages.all.all.uncompressedData": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total uncompressed message size (excluding any compression, e.g. delta compression: https://ably.com/documentation/realtime/channels/channel-parameters/deltas)."
+          "description": "Total uncompressed message size (excluding any compression, e.g. delta compression: https://ably.com/docs/channels/options/deltas)."
```
m-hulbert, line 40: We say "e.g. delta compression" quite frequently, but do we actually do any other type of compression?

json-schemas/src/account-stats.json
```diff
@@ -27,17 +27,27 @@
...
+        },
+        "messages.all.all.failed": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as rejected by an external integration target, or a service issue on Ably's side)"
+        },
+        "messages.all.all.refused": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of messages that were refused by Ably (eg due to a rate limit, a malformed message, or incorrect permissions on the client's part)"
```
m-hulbert, line 50: "eg" --> "e.g." for consistency throughout. "For example" if we want to match with the style guide.

json-schemas/src/account-stats.json
```diff
@@ -97,7 +137,17 @@
         "messages.inbound.realtime.messages.uncompressedData": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total uncompressed inbound realtime message size (excluding any compression, e.g. delta compression: https://ably.com/documentation/realtime/channels/channel-parameters/deltas, received by the Ably service from clients), excluding presence messages."
+          "description": "Total uncompressed inbound realtime message size (excluding any compression, e.g. delta compression: https://ably.com/docs/channels/options/deltas, received by the Ably service from clients), excluding presence messages."
+        },
+        "messages.inbound.realtime.messages.failed": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of inbound realtime messages excluding presence messages that failed (which did not succeed for some reason other than Ably actively rejecting it, such as ejected by an external integration target, or a service issue on Ably's side)"
```
m-hulbert, line 145: "ejected" --> "rejected"

json-schemas/src/account-stats.json
```diff
@@ -607,52 +922,22 @@
...
           "inclusiveMinimum": 0,
-          "description": "Total processed presence message count based on configured channel rules."
+          "description": "Total number of message deltas successfully generated (using xdelta, currently the only supported codec), see https://ably.com/docs/channels/options/deltas."
         },
-        "messages.processed.presence.data": {
+        "messages.processed.delta.xdelta.skipped": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total processed presence message size based on configured channel rules."
+          "description": "Total number of messages on channels where delta generation was enabled but where we skipped delta generation, eg because the payload was too small to make it worthwhile, see https://ably.com/docs/channels/options/deltas."
```
m-hulbert, line 935: "we" --> "Ably" to avoid ambiguity

json-schemas/src/account-stats.json
```diff
@@ -902,27 +1187,27 @@
...
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Peak rate of reactor webhook invocations."
+          "description": "Peak rate of events sent to webhook configured through an integration rule."
         },
         "peakRates.pushRequests": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Peak rate of pushRequests."
+          "description": "Peak rate of push api requests."
```
m-hulbert, line 1210: "api" --> "API"

m-hulbert (commented):
Just a couple of consistency updates and suggestions that appear a few times throughout.

json-schemas/src/account-stats.json
paddybyers: Personally I think the style guide is wrong and we should be using open punctuation in this respect. I'd prefer that we use `eg`, `ie` etc. `e.g.` IMO is a bit last-century.

m-hulbert: Just to clarify; I meant that in the schema we have a mix of `eg` and `e.g.`, so should opt for one or the other to be consistent (I don't disagree with `eg` being preferable at all, but there were more instances of `e.g.` already).

I'm in the same camp as the style guide though and prefer `for example`.

json-schemas/src/account-stats.json
```diff
@@ -27,17 +27,27 @@
         "messages.all.all.count": {
           "type": "number",
           "inclusiveMinimum": 0,
-          "description": "Total message count."
+          "description": "Total number of successful messages, summed over all message types and transports."
```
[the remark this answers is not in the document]
SimonWoolf: This is the `all` direction (vs `inbound` or `outbound`), so both

json-schemas/src/account-stats.json
SimonWoolf: I've reworded

json-schemas/src/account-stats.json
SimonWoolf: I've replaced with `eg`

PR #256, opened 2024-07-26
JSON schemas: bump version of stats

SimonWoolf:
About half a second after I merged https://github.com/ably/ably-common/pull/255 I realised I forgot to bump the schema version

### DIFF
--- a/json-schemas/versions.json
+++ b/json-schemas/versions.json
@@ -1,7 +1,7 @@
 {
   "agents": "0.0.2",
-  "account-stats": "0.0.2",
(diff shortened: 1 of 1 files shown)


SimonWoolf:
(at some point we should probably bump it to 1.0.0 :thinking:)

SimonWoolf:
(also we should upload to a url without the patch version too, so the schema url we send in the stats doesn't have to be updated when we make fixes & wording tweaks that don't change the payload)

m-hulbert (approved):

PR #257, opened 2024-08-01
Add unbatched outbound stats

JoaoDiasAbly:
https://ably.atlassian.net/browse/REA-1946

Adding unbatched outbound stats counts to have a comparison point with the batched results. If batching is disabled, it should contain the same value as the simple `.count` counterpart stat.

### DIFF
--- a/json-schemas/src/app-stats.json
+++ b/json-schemas/src/app-stats.json
@@ -93,12 +93,12 @@
           "inclusiveMinimum": 0,
           "description": "Total uncompressed presence message size, excluding delta compression https://ably.com/docs/channels/options/deltas."
         },
-        "messages.all.messages.failed": {
+        "messages.all.presence.failed": {
           "type": "number",
           "inclusiveMinimum": 0,
           "description": "Total number of presence messages excluding presence messages that failed (which did not succeed for some reason other than Ably explicitly refusing it, such as rejected by an external integration target, or a service issue on Ably's side)"
         },
-        "messages.all.messages.refused": {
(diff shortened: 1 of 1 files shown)


SimonWoolf:
> If batching is disabled, it should contain the same value as the simple .count counterpart stat.

That's not what the implementation in this PR does though.

AFAICT right now this PR is adding an `unbatchedCount` which is just the raw number of Messages, `messages.length`, as measured outbound by the attachment. (i.e. the old meaning of messages.count before that was changed to ProtocolMessages). That will differ from the simple `.count` even with batching disabled, whenever someone publishes a ChannelMessage with multiple Messages even without batching enabled.

We could implement this by adding the stat in the channel using the subscribe occupancy, though it will be an estimate. But from https://ably.atlassian.net/browse/REA-1946?focusedCommentId=60044 it looks like the plan was to add the batching stats to `processed` (eg `processed.batched.inbound`, `processed.batched.outbound`, maybe `processed.batched.outboundFanout` & `processed.batched.outboundFanoutUnbatched` or whatever) - is there a discussion somewhere about changing that that you could link?

SimonWoolf (commented):
Note that if we use this approach this stat will need to be added to all the stats that this aggregates into (not just messages.outbound.realtime.messages.unbatchedCount, but also messages.outbound.realtime.all.unbatchedCount, messages.outbound.all.all.unbatchedCount, messages.all.all.unbatchedCount, plus the presence ones)

but let's discuss the approach first today

PR #263, opened 2024-09-19
channel-state: Add channel state error codes and messages

kaschula:
Using the error code range 92000 add all the error messages for channel state.


protocol/errors.json
```diff
@@ -182,6 +182,39 @@
   "91004": "unable to automatically re-enter presence channel",
   "91005": "presence state is out of sync",
   "91100": "member implicitly left presence channel (connection closed)",
+  "92000": "Invalid MapOp: key required",
+  "92001": "Channel state max objects limit exceeded",
+  "92002": "Channel message must contain state operation message",
+  "92003": "Invalid state message object ID length",
```
mschristensen, line 188: Do you think we need to be this granular or is it worth having a single `Invalid state operation` error and we add additional context in the error returned to the client?

protocol/errors.json
```diff
@@ -182,6 +182,39 @@
...
+  "92016": "Invalid operation: expected a create operation",
+  "92017": "Invalid map operation: MapOp should not be provided",
+  "92018": "Invalid map operation: MapOp should be provided",
+  "92019": "Invalid map operation: Data required",
+  "92020": "Invalid map operation: unknown action",
+  "92021": "Invalid counter operation: CounterOp should be provided",
+  "92022": "Invalid counter operation: CounterOp should have non zero value:",
+  "92023": "Invalid counter operation: Counter should not be provided",
+  "92024": "State data required",
+  "92025": "Invalid State Data: both data and object id can not be set",
```
mschristensen, line 210:
```suggestion
  "92025": "Invalid state data: both data and object id can not be set",
```

protocol/errors.json
```diff
@@ -182,6 +182,39 @@
...
+  "92026": "Invalid State Data: either data or object id must be set",
+  "92027": "State object required",
+  "92028": "Invalid state object: a map or counter required",
+  "92029": "State value required",
+  "92030": "Invalid counter operation: unknown action",
+  "92031": "Channel message required",
```
mschristensen, line 216: Is this a state-specific error? I can't find a reference to it in the code

protocol/errors.json
```diff
@@ -182,6 +182,39 @@
...
+  "92032": "Invalid channel message type: state type require",
```
mschristensen, line 217:
```suggestion
  "92032": "Invalid channel message type: state type required",
```

mschristensen: also can't find a reference to this error message in the code

mschristensen (changes requested):

protocol/errors.json
kaschula: It is used in the LiveObjectPool when checking the action type on the channel message

protocol/errors.json
```diff
@@ -182,6 +182,18 @@
   "91004": "unable to automatically re-enter presence channel",
   "91005": "presence state is out of sync",
   "91100": "member implicitly left presence channel (connection closed)",
+  "92000": "Invalid message type: state type required",
+  "92001": "Channel state max objects limit exceeded",
+  "92002": "State messages required",
+  "92003": "State message required",
```
mschristensen, line 188: Can we unify this into a single code, `State message(s) required`?

kaschula: Yeah make sense

mschristensen (approved):

PR #265, opened 2024-10-14
Add error 50305 to errors.json

JoaoDiasAbly:
https://ably.atlassian.net/browse/REA-2121

Created another ticket to add to `errorsHelp.json` later on when FAQ page exists

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -107,6 +107,7 @@
 
   "50010": "Ably's edge proxy service has encountered an unknown internal error whilst processing the request",
   "50210": "Ably's edge proxy service received an invalid (bad gateway) response from the Ably platform",
+  "50305": "Ably's routing layer was unable to service this request",
   "50310": "Ably's edge proxy service received a service unavailable response code from the Ably platform",
 
   "50320": "Active Traffic Management: traffic for this cluster is being temporarily redirected to a backup service",


protocol/errors.json
```diff
@@ -107,6 +107,7 @@
 
   "50010": "Ably's edge proxy service has encountered an unknown internal error whilst processing the request",
   "50210": "Ably's edge proxy service received an invalid (bad gateway) response from the Ably platform",
+  "50305": "Ably's routing layer was unable to service this request",
```
lmars, line 110: Let's add 50205 and 50405 too, which are also set by the router in a similar place to 50305.

lmars:
@JoaoDiasAbly I know this is an old one, but if we still think it's something we want to merge, it needs to be updated to now register the codes by [creating a Markdown file in the errors/codes directory](https://github.com/ably/ably-common/tree/main/errors#error-documentation).

protocol/errors.json
JoaoDiasAbly: Done — 50205 and 50405 are now included, along with 50306 which is raised in the same `serveError` switch and was also unregistered.

JoaoDiasAbly:
Reworked to the new registration process: the four codes are now registered as Markdown files in `errors/codes/` and `protocol/errors.json` is regenerated from the registry (no hand edits). Branch was reset onto `main`, so the conflict is gone. Updated the PR description with the details.

errors/codes/50205.md
lmars, line 1: Do you mind giving Claude a stab at writing the bodies too? I'm currently working on the bodies for all existing codes

lmars: PR where I've added a bunch of body text: https://github.com/ably/ably-common/pull/352

JoaoDiasAbly: done

errors/codes/50306.md
```diff
@@ -0,0 +1,18 @@
+---
+code: 50306
+identifier: region_temporarily_disabled
+title: Region temporarily disabled
+summary: The request reached a region that has been temporarily taken out of service. This normally affects requests made to a region-specific endpoint while that region is disabled.
+---
+
+## What you should do
+
+On the main endpoint (the default), usually nothing: traffic is directed away from a disabled region, and SDK retries reach a healthy one. On a [region-specific endpoint](https://ably.com/docs/platform/account/enterprise-customization#regional-routing-and-storage), switch to the main endpoint: that endpoint pins every attempt to the named region, including retries, so requests keep failing while it is disabled. If your application has to stay pinned to a region, [contact Ably support](https://ably.com/support).
```
lmars: This is bad advice, if the customer is using a region-specific endpoint it's typically because they have a regional placement constraint, so if they switch to the main endpoint and hit regions their account is not permitted in, then they'll get errors.

Also, customers don't "pin to a region", the regional routing pins to a geographic area spanning multiple regions, so the advice here should be similar to 50305, the SDK will retry and traffic will eventually stop hitting the disabled region (as we would have taken it out of DNS).

errors/codes/50306.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Why it happens
+
+Ably occasionally takes a region out of service temporarily, for example to protect the platform during an incident. A disabled region rejects the requests that reach it rather than redirecting them elsewhere.
```
lmars: It's never to protect the platform, it's always to prevent disruption when services in a region are not functioning correctly, often due to reasons outside out control.

errors/codes/50306.md
JoaoDiasAbly: I see... That makes sense, thanks for calling it out. Fixed now

errors/codes/50306.md
JoaoDiasAbly: Overlooked that wording, thanks. Fixed

PR #272, opened 2024-10-26
errors: simplify channel state error codes

mschristensen:
There were many error codes for different variants of invalid state message errors. This will quickly become unwieldy and I think it's better if these errors instead share a single code. Additional context can be included on the error message for specific errors.

Error codes each have their own FAQ page. I think we're better off with a single FAQ page for invalid state messages that describes what a valid message is in different contexts.

This is breaking change, but the channel state work that makes use of these codes has not yet been released so it is safe to make the change now.


kaschula:
Hi Mike, I was going to suggest that the result of this is we will probably end up with on large FAQ page covering a range of errors to do with different operations and requirements they have. After reading the PR description I can see that this is the intent. So happy to go ahead. I think given most of these errors are validation related we can be quite concise about what the different requests should look like.

kaschula (approved):

PR #273, opened 2024-10-31
Update account-stats and app-stats schemas for High Throughput Pricing

JoaoDiasAbly:
https://ably.atlassian.net/browse/REA-2157


json-schemas/src/account-stats.json
```diff
@@ -339,6 +359,11 @@
           "inclusiveMinimum": 0,
           "description": "Total outbound realtime message count (sent from the Ably service to clients), excluding presence messages."
         },
+        "messages.outbound.realtime.messages.billableCount": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of outbound realtime messages that are billable (sent from the Ably service to clients), excluding presence messages."
```
mschristensen, line 365: does this exclude state messages too?

json-schemas/src/account-stats.json
```diff
@@ -859,6 +899,11 @@
           "inclusiveMinimum": 0,
           "description": "Total outbound presence message count (sent from the Ably service to clients)."
         },
+        "messages.outbound.all.presence.billableCount": {
```
mschristensen, line 902: Do we need an equivalent for state messages?

json-schemas/src/account-stats.json
JoaoDiasAbly: do state messages get counted in `messages.outbound.realtime.messages.count`?

json-schemas/src/account-stats.json
JoaoDiasAbly: As far as I can tell there are no state specific stats entries in the file. I'd say we probably want that when we add server-side-batching for state messages (assuming we'll want that in the future).

How are state message stats currently counted? Are they counted as regular messages?

json-schemas/src/account-stats.json
mschristensen: no they have their own key separate from regular messages, exactly like presence

json-schemas/src/account-stats.json
mschristensen: We do have state stats counted, but these schemas haven't been updated - I've made a note to fix that. They are counted separately from regular messages, and if `billableCount` is batching-related only, then yes I don't think we need state equivalents specified here

mschristensen (approved):

json-schemas/src/account-stats.json
JoaoDiasAbly: Is and isn't; AFAICT this will be the new field used for billing customers, but the way it's populated will differ whether batching is enabled or not. I've actually already included the correct counting for them here https://github.com/ably/realtime/pull/6724 (which makes my previous question sound a bit silly since I actually knew the answer - that we do have state stats counted - but wasn't remembering because this ticket only covers the request in https://github.com/ably/realtime/issues/6585)

to give you further context: for messages and presence outbound (and state, even though we don't have batching for it rn) we count the batches we see in the billableCount; if batching is disabled then billableCount == count

PR #275, opened 2024-11-07
adds error code faqs for mongodb connector

jamienewcomb:
Error Code 72004 - "Ingress cannot identify channel, no _ablyChannel field"
Error Code 72005 - "Ingress invalid pipeline"
Error Code 72006 - "Unable to resume from change stream"
Error Code 72007 - "Unable to store change stream resume token"


zknill (approved):
brill! 👍 

I opened every link and checked those too -- thanks!

PR #280, opened 2024-12-02
errors: cannot submit op to tombstone

mschristensen:
Adds a channel state error code returned to the client when an operation is submitted to an object that is marked as a tombstone.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -195,6 +195,7 @@
 
   "92000": "invalid state message",
   "92001": "state max objects limit",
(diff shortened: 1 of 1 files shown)


protocol/errors.json
```diff
@@ -195,6 +195,7 @@
 
   "92000": "invalid state message",
   "92001": "state max objects limit",
+  "92002": "cannot submit operation on tombstone object",
```
zknill, line 198: nit: "unable to"?

zknill (approved):

PR #288, opened 2025-01-13
feat: DTP-1126 Update App and Account stats schemas with state messages

kaschula:
This PR adds REST stat schema entries, both inbound and outbound for statemessages.

It also adds missing regular message entires and updates the description for the regular message to now say that presence and state messages are excluded from those stats.


owenpearson (approved):
LGTM

PR #289, opened 2025-01-15
Add new batching and conflation processed stats

JoaoDiasAbly:
Ticket: https://ably.atlassian.net/browse/REA-2258

Related PR: https://github.com/ably/realtime/pull/7027


SimonWoolf:
I've replied on the realtime pr with comments about what the terminology, so will leave discussion on that to there.

(one review comment that I didn't put there but would here is the use of "message" here to mean a ChannelMessage, which is not what message means elsewhere in stats)

json-schemas/src/account-stats.json
```diff
@@ -1097,6 +1094,36 @@
           "inclusiveMinimum": 0,
           "description": "Total number of messages on channels where delta generation was attempted but where a delta was not successfully generated, so Ably just broadcasted the original, see https://ably.com/docs/channels/options/deltas."
         },
+        "messages.processed.batching.simple.inboundMessages": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of messages going through the batching pipeline with the 'simple' batching policy."
```
SimonWoolf, line 1100: wondering if "going through the batching pipeline" might be a little bit too ably-internal-jargon. maybe just "aggregated into batches" (or "aggregated into batches before conflation" for the conflation type)? or "published on channels with batching enabled"?

json-schemas/src/account-stats.json
```diff
@@ -1097,6 +1094,36 @@
...
+        },
+        "messages.processed.batching.simple.generatedBatches": {
+          "type": "number",
+          "inclusiveMinimum": 0,
+          "description": "Total number of batches generated by the 'simple' batching policy."
+        },
+        "messages.processed.batching.simple.discardedMessages": {
```
SimonWoolf, line 1107: given this should always be 0, do we need to include it? in other places we haven't bothered to include entries for stats which will never be non-zero

SimonWoolf (approved):
couple of comments, but generally lgtm!

PR #293, opened 2025-03-12
Add error codes for live object rest apis

zknill:
These two error codes are used for fetching live object state from the rest api. 
They indicate two cases where the API is unable to provide a response because of the state of the object.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -198,6 +198,11 @@
   "92000": "invalid object message",
   "92001": "objects limit exceeded",
   "92002": "unable to submit operation on tombstone object",
+  "92003": "unable to fetch object tree with tombstone object as root",
+  "92004": "object not found",
+  "92005": "no objects found matching operation path",
+  "92006": "unable to perform operation without objectId or path",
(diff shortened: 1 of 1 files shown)


protocol/errors.json
```diff
@@ -198,6 +198,8 @@
   "92000": "invalid object message",
   "92001": "objects limit exceeded",
   "92002": "unable to submit operation on tombstone object",
+  "92003": "unable to fetch state tree with tombstone object as root",
```
mschristensen, line 201:
```suggestion
  "92003": "unable to fetch objects tree with tombstone object as root",
```

protocol/errors.json
```diff
@@ -198,6 +198,8 @@
...
+  "92004": "unable to fetch state tree for not found object as root",
```
mschristensen, line 202:
```suggestion
  "92004": "object not found",
```

per suggestion in https://github.com/ably/realtime/pull/7212#discussion_r1993943642

mschristensen (approved):

PR #300, opened 2025-05-07
Update datacenter-locations.csv to include Melbourne

mattheworiordan:
I noticed our Ably demo (ably/demo) had one missing DC, but then noticed this is because it was based off the locations in this file.


paddybyers (approved):
LGTM, thanks

PR #302, opened 2025-05-18
Add support for ably-cli in agents header

mattheworiordan:
Note that describing the ably-cli as a wrapper is arguably not correct, it's an application or perhaps a tool, but adding support for a new type is probably a lot of work, and at this stage, has limited value given we simply want a way to track ably-cli usage. The reason it's not correct is in the context of the data plane, ably-cli is not providing an API to end users, and wrapping functionality in an SDK, it is just a tool or app consuming. When it's calling the control plane, then it's not really an SDK as it's just a tool calling the API directly, and it's not a wrapper as it's not wrapping anything. If I categorised ably-cli as an SDK, then we'd also have the problem that the underlying SDK would not be logged as I believe only one SDK value is supported in most reports. On balance, I think what I have chosen is correct, but open to feedback / suggestions.

This commit also adds instructions for future people who want to make changes to agents to make them aware that agents are duplicated in our analytics pipeline processing.

Once this PR is merged, two downstream actions are needed:

1) router to be updated and deployed
2) analytics pipeline to be updated to support the new ably-cli agent.

### DIFF
--- a/protocol/README.md
+++ b/protocol/README.md
@@ -93,6 +93,7 @@ The following table adds more contextual detail for some agent identifiers, wher
 | ---------- | -------- |
 | `ably-asset-tracking-android` | [ably-asset-tracking-android](https://github.com/ably/ably-asset-tracking-android) |
 | `ably-asset-tracking-swift` | [ably-asset-tracking-swift](https://github.com/ably/ably-asset-tracking-swift) |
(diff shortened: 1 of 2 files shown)


lmars (approved):

mattheworiordan:
Thanks @lmars

PR #304, opened 2025-05-23
errors: add new AMPQ error codes

kaschula:
Add 2 new 500 status error codes for when the AMQP message delivery fails.


protocol/errors.json
```diff
@@ -125,7 +125,8 @@
   "70002": "reactor operation failed (post operation returned unexpected code)",
   "70003": "reactor operation failed (maximum number of concurrent in-flight requests exceeded)",
   "70004": "reactor operation failed (invalid or unaccepted message contents)",
-
+  "70005": "unable to deliver AMQP message (unspecified timeout)",
+  "70006": "unable to deliver AMQP message (cluster busy)",
```
amnonbc, line 129: maybe keep the empty line before 71000 ?

amnonbc (approved):
minor comment
otherwise LGTM

PR #307, opened 2025-06-16
errors: update annotation error messages

mschristensen:
Avoid references to "namespace" which is not terminology we use externally.

Removes 93003 which is a duplicate of 93002 and is unused in Realtime and in SDKs.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -206,8 +206,8 @@
   "92006": "unable to perform operation without objectId or path",
   "92007": "operation not processable on path",
 
-  "93001": "attempt to add an annotation listener without having requested the annotation_subscribe channel mode in ChannelOptions, which won't do anything (we only deliver annotations to clients who have explicitly requested them)",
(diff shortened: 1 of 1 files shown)


lmars:
@mschristensen we recently [updated how error codes are maintained in this repo](https://github.com/ably/ably-common/pull/348), and as part of that, the title and summary of the codes changed in this PR are now:
- `93001` - **Annotation listener without annotation_subscribe mode**
  An annotation listener was added to a channel that had not requested the annotation_subscribe mode in its channel options. Annotations are only delivered to clients that explicitly request them, so the listener will not receive anything.
- `93002` - **Message annotations, updates, appends, and deletes not enabled**
  An operation could not be performed because the channel did not have the 'Message annotations, updates, appends, and deletes' rule enabled.

Given these don't use the "namespace" terminology, and there is no 93003 error code in the registry, I'm going to assume this PR is no longer needed.

PR #313, opened 2025-07-21
Add `chat-ui-kit` entry to protocol documentation and agents list

splindsay-92:
With the release of the [chat-ui-kit](https://github.com/ably/ably-chat-react-ui-components), we've added an agent string to track usage. This needs to be added in common so we can have tracking in metabase.

[CHA-1062](https://ably.atlassian.net/browse/CHA-1062)


protocol/agents.json
```diff
@@ -365,6 +365,11 @@
       "versioned": true,
       "type": "wrapper"
     },
+    {
+      "identifier": "chat-ui-kit",
```
AndyTWF, line 369: I would suggest `chat-react-ui-components` to match the repo name and ensure we don't have a clash later if we introduce new kits for Kotlin/Swift

AndyTWF: Yeah

AndyTWF (approved):

PR #317, opened 2025-08-24
Improve agent information to include SDK info + releases

mattheworiordan:
This PR enables us to deliver useful SDK reporting in our dashboards by pushing more understandable agent information into agents.json and adding release information that is determined from every SDK's tags + our deprecation policy at https://ably.com/docs/platform/deprecate.

See new SDK report at https://ably-metabase.herokuapp.com/dashboard/1396-account-sdk-usage-dashboard enabled through having this data.


.env.example
```diff
@@ -0,0 +1,15 @@
+# GitHub API Token for fetching releases (optional but recommended for rate limits)
+GITHUB_TOKEN=your_github_token_here
+
+# OpenAI API Key for summarizing release notes (optional)
+OPENAI_API_KEY=your_openai_api_key_here
+
+# AWS Configuration for S3 uploads
+# Note: AWS_ROLE_ARN and AWS_REGION are used in GitHub Actions CI only (set as repository secrets)
+# For local development, use AWS CLI profiles or environment variables
+S3_BUCKET=files.ably.io
```
emptyhammond, line 10: Can we upload to `ably-analytics-data-lake-production`, we have existing CSV import pipelines that we can use to ingest this data, but these rely on the data being present in this location.

I will need to provide you with an alternative `AWS_ROLE_ARN` that has permissions to write to this bucket.

emptyhammond (commented):
Looks good, but we need to update the targets bucket to leverage the existing CSV import pattern/pipeline. 

The Snowflake import steps can ultimately be deprecated as they will be superseded by a pipeline implemented in the infra repo.

.env.example
mattheworiordan: Sure, happy to do this, or just issue a change on this PR if you want. I have no preference where it should go, just that it needs to be in S3 somewhere, and I expect @zychua25 would make this work anyway when he sets up imports into Snowflake.

.github/workflows/sync-to-s3.yml
```diff
@@ -0,0 +1,106 @@
...
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          role-session-name: GitHubActions-ably-common
+          aws-region: ${{ secrets.AWS_REGION }}
+
+      - name: Sync CSV files to S3
+        run: |
+          echo "🔄 Syncing CSV files to S3..."
+          echo "   Bucket: $S3_BUCKET"
```
emptyhammond, line 40:
```suggestion
          echo "   Bucket: ${{ vars.S3_BUCKET_NAME || env.S3_BUCKET }}"
```

.env.example
emptyhammond: Ok, I've pushed two commits to remove the sql related code as this will be handled by a data pipeline elsewhere and externalise the S3 config in the action.

scripts/export-agents-csv.js
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+
+/**
+ * Exports agents.json to CSV format for easy database import
```
lmars, line 7: What database are these being imported into which doesn't support JSON?

scripts/export-agents-csv.js
```diff
@@ -0,0 +1,87 @@
...
+  const type = agent.type || '';
+  const source = agent.source || '';
+  const product = agent.product || '';
+  const name = agent.name || '';
+  
+  // Escape fields that might contain commas
+  const escapeCsvField = (field) => {
+    const str = String(field);
+    if (str.includes(',') || str.includes('"') || str.includes('\n')) {
+      return `"${str.replace(/"/g, '""')}"`;
```
lmars, line 35: Why is this enclosed in backticks?

scripts/fetch-agent-releases.js
```diff
@@ -0,0 +1,616 @@
...
+ */
+function getReleaseType(version) {
+  const normalized = version.toLowerCase();
+  
+  if (normalized.includes('-rc') || normalized.includes('.rc') || normalized.includes('+rc') || 
+      normalized.includes('release-candidate') || normalized.includes('-preview') || normalized.includes('+preview')) {
+    return 'release-candidate';
+  }
+  if (normalized.includes('-beta') || normalized.includes('.beta') || normalized.includes('+beta') || 
+      normalized.includes('-swift') || normalized.includes('+swift')) {
```
lmars, line 310: Is this categorisation based on a convention we actually use?

scripts/fetch-agent-releases.js
```diff
@@ -0,0 +1,616 @@
...
+
+/**
+ * Check if a version is less than or equal to a threshold
+ */
+function isVersionLessThanOrEqual(version, threshold) {
+  return compareVersions(version, threshold) <= 0;
+}
+
+/**
+ * Calculate support end date (12 months after release)
```
lmars, line 363: Is this our standard policy? It seems rather arbitrary.

scripts/fetch-agent-releases.js
```diff
@@ -0,0 +1,616 @@
...
+ */
+function calculateSupportEndDate(releaseDate) {
+  const date = new Date(releaseDate);
+  date.setFullYear(date.getFullYear() + 1);
+  return date.toISOString().split('T')[0];
+}
+
+/**
+ * Calculate deprecation date (24 months after release)
```
lmars, line 372: Similar to above, is this standard policy? I don't think a release becomes deprecated just because it's 24 months old, surely it becomes deprecated because it relies on deprecated functionality that will be removed in the near future?

lmars (commented):
It was mentioned that this is blocked on an engineering review so I offered to take a look, I've left a few comments but let me know if you want me to help work to resolve them.

scripts/fetch-agent-releases.js
paddybyers: See https://ably.com/docs/platform/deprecate#timelines

lmars: Ok thanks, so then this is incorrect, since it's based on the date of the release in question, not based on the date of the release which supersedes it?

paddybyers: Correct

.env.example
mattheworiordan: Thanks @emptyhammond

.github/workflows/sync-to-s3.yml
```diff
@@ -0,0 +1,106 @@
...
+            --delete
+          
+          echo "✅ Sync complete!"
+          
+          # List uploaded files
+          echo ""
+          echo "📋 Uploaded files:"
+          aws s3 ls s3://${S3_BUCKET}/${S3_PREFIX}/ --recursive
+
+  generate-and-sync:
```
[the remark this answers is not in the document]
mattheworiordan: Good spot, will fix, I assume just a hangover from when this was just a fledgling idea.

.github/workflows/sync-to-s3.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Generate release data
+        run: |
+          # Only run if we have the required tokens
+          if [ -n "${{ secrets.GITHUB_TOKEN }}" ] && [ -n "${{ secrets.OPENAI_API_KEY }}" ]; then
+            echo "Generating release data..."
+            GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} \
+            OPENAI_API_KEY=${{ secrets.OPENAI_API_KEY }} \
+            npm run fetch:agent-releases
+          else
+            echo "⚠️  Skipping release data generation (missing tokens)"
```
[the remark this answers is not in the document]
mattheworiordan: OK

data/agents/README.md
```diff
@@ -0,0 +1,32 @@
+# Ably Agent Data
+
+This directory contains CSV exports of Ably agent metadata for analysis and database import.
```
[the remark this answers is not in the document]
mattheworiordan: Well at the time I wanted them for two treasons:

- We have a Git record of them.
- Other tools like ably-os will have access to these files for analysis.

Saying that, I recognise this introduces a new issue in that the committed version could fall out of sync with the S3 version (given that job runs daily), and that's not ideal, and I don't want to change the CI job to auto-commit changes. 

So on balance, I agree we should remove this, and update this README to link to the S3 files, so long as they are publicly available on that URL. I will check with @emptyhammond if that will be possible for the target S3 repo he has suggested this is uploaded to.

package.json
```diff
@@ -8,7 +8,12 @@
     "format:data": "prettier --write '**/*.{json,yml}'",
     "format:data:check": "prettier --check '**/*.{json,yml}'",
     "format:documentation:check": "markdownlint-cli2 \"**/*.md\" \"#node_modules\"",
-    "test": "jest"
+    "test": "jest",
+    "validate:agents": "node scripts/validate-json-schema.js protocol/agents.json",
+    "validate:json-schema": "node scripts/validate-json-schema.js",
+    "fetch:agent-releases": "node scripts/fetch-agent-releases.js",
+    "export:agents": "node scripts/export-agents-csv.js",
+    "upload:s3": "./scripts/upload-to-s3.sh"
```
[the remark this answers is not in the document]
mattheworiordan: Personally I like it because there is no way to push a change to S3 locally otherwise. Is this something you don't think we should support?

protocol/agents.json
```diff
@@ -1,379 +1,659 @@
...
+            {
+              "upToVersion": "1.2.35",
+              "sunsetDate": "2025-11-01",
+              "reason": "Protocol v2 migration - SDK versions that do not support Ably protocol v2 will no longer be supported"
+            }
+          ]
+        }
+      ],
+      "source": "https://github.com/ably/ably-js",
+      "product": "pub-sub",
```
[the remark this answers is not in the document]
mattheworiordan: Yes, they are considered different products https://ably.atlassian.net/wiki/spaces/product/pages/3079569421/Our+products+and+how+they+stack+up+within+our+multi-product+offering.

> I'm wondering if this should be products for the case where a given SDK can be used with multiple products

I am not sure I agree. The ably-js is part of the PubSub product, and a plugin exists to use LiveObjects from PubSub, that does not mean ably-js is LiveObjects. LiveObjects can also be used with Spaces or Chat, but we wouldn't tag those SDKs as LiveObjects.

Note Push Notifcations right now is not considered a product, but we have discussed productising it at some point.

scripts/export-agents-csv.js
mattheworiordan: Snowflake.

package.json
```diff
@@ -8,7 +8,12 @@
     "format:data": "prettier --write '**/*.{json,yml}'",
     "format:data:check": "prettier --check '**/*.{json,yml}'",
     "format:documentation:check": "markdownlint-cli2 \"**/*.md\" \"#node_modules\"",
-    "test": "jest"
+    "test": "jest",
+    "validate:agents": "node scripts/validate-json-schema.js protocol/agents.json",
+    "validate:json-schema": "node scripts/validate-json-schema.js",
```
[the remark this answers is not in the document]
mattheworiordan: So that you can use the tool to validate any JSON file, so for example:

```
$ npm run validate:json-schema protocol/errors.json

> ably-common@1.2.0 validate:json-schema
> node scripts/validate-json-schema.js protocol/errors.json

❌ Error: No schema provided and JSON file does not contain $schema property
   Please provide a schema path/URL as the second argument
```

I think this is useful. Would you like to me to perhaps improve naming (annoying that npm package.json doesn't allow descriptions) such as `validate:json-schema:agents` for the agents file, and `validate:json-schema` for the command where you provide a path?

data/agents/README.md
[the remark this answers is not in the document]
lmars: > I don't want to change the CI job to auto-commit changes

As an FYI we do have an existing GitHub Action which generates Go code and commits back to GitHub:

https://github.com/ably/ably-common/blob/783496f4dd1110d0175a7ced60de826e1aba40cc/.github/workflows/publish.yml#L24-L38

Although it does commit to [ably-common-go](https://github.com/ably/ably-common-go) which we have explicitly stated contains auto-generated code.

So if a git based workflow is preferred, perhaps let's do the same here and commit to, say, ably/ably-agents or something?

> so long as they are publicly available on that URL

I suspect we don't want to allow any public files in our prod analytics data lake bucket, I think we'd want to put them in a separate bucket tbh.

package.json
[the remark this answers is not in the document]
lmars: Ok fair enough, happy to leave as is 👍

protocol/agents.json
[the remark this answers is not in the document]
lmars: Ok, I guess I just don't understand the product structure enough so I defer to your judgement, but the new dashboard does have Push Notifications listed as a product, is that incorrect?

Screenshot 2025-10-06 at 09 57 21

data/agents/README.md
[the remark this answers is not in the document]
emptyhammond: >  I will check with @emptyhammond if that will be possible for the target S3 repo he has suggested this is uploaded to.

No they wouldn't be so we'd need to work around that, probably by walking back where we publish the files and syncing over to where we need them internally.

scripts/export-agents-csv.js
lmars: Ah I didn't realise this is returning a quoted string, thanks.

scripts/fetch-agent-releases.js
lmars: > I based it on what actually exists in our tags

Ok, I just wanted to check this wasn't the LLM picking some arbitrary tags, might be worth a comment that these tags are based on analysing all the tags we actually use.

scripts/fetch-agent-releases.js
```diff
@@ -0,0 +1,617 @@
...
+ * Check if a version is less than or equal to a threshold
+ */
+function isVersionLessThanOrEqual(version, threshold) {
+  return compareVersions(version, threshold) <= 0;
+}
+
+/**
+ * Calculate support end date (12 months after release)
+ */
+function calculateSupportEndDate(releaseDate) {
```
lmars, line 365: @mattheworiordan I think this also needs to be based on the superseding release date.

data/agents/README.md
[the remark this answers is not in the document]
mattheworiordan: @emptyhammond ok given you changed this to push these files to that repo, can you make a change/suggest an alternative so that I can address that?

@lmars I am tempted to not to go down the Git commit route. I don't think we need git commit history in practice, and it just adds more complexity and repos, for little additional value IMHO.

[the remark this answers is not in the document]
mattheworiordan: > But to my original point, should this data directory be deleted now?

Yes, but I was waiting on confirmation from @emptyhammond as to where I can push these files.

PR #331, opened 2026-03-24
Add error codes 104000-104005 for AI Transport SDK

mschristensen:


lmars:
@mschristensen we recently [updated how error codes are maintained in this repo](https://github.com/ably/ably-common/pull/348), so if these new codes are still needed, they need to be registered by [creating a Markdown file in the errors/codes directory](https://github.com/ably/ably-common/tree/main/errors#error-documentation).

PR #337, opened 2026-05-06
Add ai-transport-js to the agent registry

zknill:
Register the `ai-transport-js` wrapper agent (versioned) for the new `@ably/ai-transport` SDK at https://github.com/ably/ably-ai-transport-js. The SDK self-registers this identifier on the Realtime client's `options.agents` map at session construction so the Ably backend can attribute usage to it, in line with the pattern used by ably-chat-js.

Introduce a new `ai-transport` product category in the agents JSON schema enum to support the entry.

### DIFF
--- a/protocol/agents.json
+++ b/protocol/agents.json
@@ -672,6 +672,14 @@
       "name": "Chat React UI Kit"
     },
     {
+      "identifier": "ai-transport-js",
+      "versioned": true,
+      "type": "wrapper",
+      "source": "https://github.com/ably/ably-ai-transport-js",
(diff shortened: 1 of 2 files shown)


protocol/agents.json
```diff
@@ -623,6 +623,14 @@
       "product": "chat",
       "name": "Chat React"
     },
+    {
```
lawrence-forooghian: Probably would make more sense at the end of the list instead of in the middle of the Chat SDKs

lawrence-forooghian (approved):

PR #343, opened 2026-06-09
Add AI Transport SDK error codes (104000-104011)

zknill:
Register the AI Transport SDK's 104xxx custom error codes so SDK
validation against this registry passes. The codes cover SDK-specific
scenarios that don't map cleanly to existing entries:

- 104000-104008 — encoder recovery, transport lifecycle, channel
  continuity, stream piping (added with earlier branches).
- 104010 "run start deadline exceeded" — client send() did not
  observe x-ably-run-start within the configured deadline.
- 104011 "user prompt not found within lookup deadline" — agent's
  channel-rewind lookup for the client-published user message
  lapsed without a match.

Capability failures reuse Ably's existing 40160 ("operation not
permitted with provided capability") rather than coining a new code.

### DIFF
--- a/protocol/errors.json
+++ b/protocol/errors.json
@@ -254,5 +254,17 @@
   "103000": "unable to publish push notification",
   "103001": "unable to publish push notification, retries exhausted",
   "103002": "unable to publish push notification, no recipient for direct publish",
-  "103003": "unable to publish push notification, cannot handle body"
+  "103003": "unable to publish push notification, cannot handle body",
+
+  "104000": "encoder recovery failed after flush",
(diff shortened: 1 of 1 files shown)


lmars:
@zknill we recently [updated how error codes are maintained in this repo](https://github.com/ably/ably-common/pull/348), so if these new codes are still needed, they need to be registered by [creating a Markdown file in the errors/codes directory](https://github.com/ably/ably-common/tree/main/errors#error-documentation).

Also worth noting that this looks to be a dupe of https://github.com/ably/ably-common/pull/331 (cc @mschristensen).

